PR #4, opened 2020-03-21
download-artifact: paginate workflow response

jonchang:
By default, the rest.js API only returns 30 results. Using `await` pagination should let us find older workflow runs.

### DIFF
--- a/.github/workflows/download-artifact-download.yml
+++ b/.github/workflows/download-artifact-download.yml
@@ -1,6 +1,7 @@
 name: Download
 
 on:
+  pull_request: []
   push:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


dawidd6:
From logs:

```
[@octokit/paginate-rest] "response.data.workflow_runs" is deprecated for "GET /repos/Homebrew/actions/actions/workflows/download-artifact-upload.yml/runs". Get the results directly from "response.data"
```

jonchang:
Yeah, I don't know where that error is coming from. I tried switching it to `response.data["workflow_runs"]` but still get the same deprecation warning.

jonchang:
Probably the thing to do here is: wait for github to update `@actions/github` to use octokit/rest.js v17, update `package.json` in a pull request, and fix the errors that pop up. The deprecation warning is not super helpful since I suspect it's interacting poorly with `registerEndpoint`.

PR #7, opened 2020-03-30
Add cancel-previous-runs

dawidd6:
This Action might be helpful for us, especially in `homebrew-core`, where users or maintainers could force push the changes multiple times in few minutes, resulting runners queue to be cluttered by old workflow runs that should be cancelled.


dawidd6:
Looks like it works.

This run: https://github.com/Homebrew/actions/actions/runs/66538853
cancelled this one: https://github.com/Homebrew/actions/actions/runs/66538264

PR #8, opened 2020-04-16
download-artifact: Fix deprecation warning

issyl0:
- I saw this in the Homebrew/linuxbrew-core Actions logs:

```
[@octokit/paginate-rest] "response.data.workflow_runs" is deprecated for "GET /repos/Homebrew/linuxbrew-core/actions/workflows/build-bottles.yml/runs". Get the results directly from "response.data"
```

### DIFF
--- a/download-artifact/main.js
+++ b/download-artifact/main.js
@@ -59,7 +59,7 @@ async function main() {
 
         let run
         for await (const response of client.paginate.iterator(options)) {
-            const matching_run = response.data.workflow_runs.find((workflow_run) => {
+            const matching_run = response.data.find((workflow_run) => {
                 return workflow_run.head_sha == commit
             })
             if (matching_run) {


issyl0:
```
##[error]response.data.find is not a function
```

Clearly I still can't write JS. Maybe that deprecation warning wasn't meant for our code? 🤔

Bo98:
Prior discussion: https://github.com/Homebrew/actions/pull/4#issuecomment-602012065

issyl0:
Ah - I'll close this and live with it then! I thought it was a new thing.

Bo98:
A possibility, since `@actions/github` seem to be slow at keeping up with Octokit, is to use `@octokit/action` instead. It does mean losing `github.context`, though I suppose that can be replicated with `const [owner, repo] = process.env.GITHUB_REPOSITORY.split('/')`.

PR #11, opened 2020-06-25
Add brew-script Action

dawidd6:
A new Action that allows us to run for example `ohai "hello"` simply. It's similar to `actions/github-script` Action, but for running Ruby scripts in Homebrew environment via `brew ruby` command.

cc @jonchang


MikeMcQuaid:
Where/how is this currently used? If it's not yet: where would it be?

I think it would be good to keep this repo to only the stuff we're using (or going to use) in our workflow files for the time being (which may include this).

dawidd6:
This could allow us to skip writing JavaScript and write Ruby instead. I mean, I know we should be making more full-fledged Actions instead of writing programming languages code in YAML, but... yea.

No need to merge it just yet.

MikeMcQuaid:
Sounds cool. What's an example you can think of that might be useful?

dawidd6:
Currently, none to be honest and don't want to force ideas.

dawidd6:
Merging as is.

PR #18, opened 2020-06-30
download-artifact: update

dawidd6:
Take recent updates from _upstream_ repo.


MikeMcQuaid:
Could we consider either:
1) making this the upstream repo
2) automating this update process

dawidd6:
Would be great to make this the upstream repo, but I have some users of the _upstream_ Action and don't want to leave them in the dark :D. I'll just try to maintain both for now.

PR #19, opened 2020-06-30
Add post-comment Action

dawidd6:
https://github.com/Homebrew/brew/issues/7805

This is just a draft for now. Still need to figure out what we exactly need in this Action and how to make it the most generic as we can.


dawidd6:
I looked at probably all of our usages of various variants of `Post comment*` steps in workflows and I concluded, that we often use a different comment body when @BrewTestBot is the `GITHUB_ACTOR`. That's why I introduced `bot_body` and `bot` inputs, so we can easily specify different comment bodies for both humans and bot separately.

Action tested locally on this PR: https://github.com/dawidd6/homebrew-test/pull/213

jonchang:
Very nice, I like the bot/user design split! This will definitely clean up a lot of the code we have for posting comments.

PR #22, opened 2020-07-04
maintain one single top-level node_modules

dawidd6:
`node` searches parent directory for `node_modules` too, so we can deduplicate some code and maintain one single `node_modules` directory with dependencies.


dawidd6:
Merging now, will make a follow-up PR with docs.

PR #25, opened 2020-07-26
Add label-pull-requests Action

dawidd6:
Per discussion here: https://github.com/Homebrew/brew/issues/7805

This PR adds an Action which labels pull requests when specified criteria match.

It makes a lot of API calls right now, cause it needs to first fetch list of open PRs, then fetch a list of changed files for each of those PRs and then fetch every changed file's content. But this way it is very flexible.

I went for JSON format as a criteria definition, as it is the easiest to implement with NodeJS, but I'm open for suggestions.
Also embedding JSON document in YAML isn't very friendly and sane, I know. Initially I thought about making a _custom DSL_ a'la GitHub search, like: `label:new formula,status:added,path:Formula/.+`, but abandoned it, as it required more work and I wanted to have a _working idea_ faster.

### DIFF
--- a/label-pull-requests/action.yml
+++ b/label-pull-requests/action.yml
@@ -15,7 +15,7 @@ inputs:
 
       It's an array of objects containing the following fields:
         - "label" (required): Label name, that should be added to PR
-        - "status" (optional): File status, one of: added, modified, changed
+        - "status" (optional): File status, one of: added, modified, removed
         - "path" (optional): Changed file path, regex matching
         - "content" (optional): Changed file content, regex matching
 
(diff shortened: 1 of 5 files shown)


label-pull-requests/action.yml
```diff
@@ -0,0 +1,36 @@
...
+    description: GitHub token
+    required: false
+    default: ${{github.token}}
+  def:
+    description: |
+      Definition in JSON format.
+
+      It's an array of objects containing the following fields:
+        - "label" (required): Label name, that should be added to PR
+        - "status" (optional): File status, one of: added, modified, changed
```
jonchang, line 18:
```suggestion
        - "status" (optional): File status, one of: added, modified, removed
```

PR #26, opened 2020-07-27
cancel-previous-runs: repurpose

dawidd6:
Repurpose existing non-used Action to sweep all queued or in progress workflow runs and cancel those previous and duplicated ones.


dawidd6:
Don't know how to test it properly here, artificially I added `sleep 9999` in workflow, then deleted it, force-pushed and new workflow run cancelled the previous one, so I guess it works.

PR #27, opened 2020-07-28
Add setup-homebrew Action

dawidd6:
This PR adds an Action which sets up Homebrew on Ubuntu or macOS.

https://github.com/Homebrew/brew/issues/7805

PR #28, opened 2020-07-29
setup-homebrew: fixes

dawidd6:
Continuation of https://github.com/Homebrew/actions/pull/27

@MikeMcQuaid Think it is all now.

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -1,13 +1,8 @@
-#/bin/bash
+#!/bin/bash
 
 set -euo pipefail
 
-if which brew &>/dev/null; then
-    HOMEBREW_PREFIX="$(brew --prefix)"
-    HOMEBREW_REPOSITORY="$HOMEBREW_PREFIX/Homebrew"
-    HOMEBREW_CORE_REPOSITORY="$HOMEBREW_REPOSITORY/Library/Taps/homebrew/homebrew-core"
-    brew update-reset "$HOMEBREW_REPOSITORY" "$HOMEBREW_CORE_REPOSITORY"
-else
+if ! which brew &>/dev/null; then
     HOMEBREW_PREFIX=/home/linuxbrew/.linuxbrew
     HOMEBREW_REPOSITORY="$HOMEBREW_PREFIX/Homebrew"
     sudo mkdir -p "$HOMEBREW_PREFIX"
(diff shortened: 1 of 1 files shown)


setup-homebrew/main.sh
```diff
@@ -1,49 +1,48 @@
-#/bin/bash
+#!/bin/bash
 
 set -euo pipefail
 
-if which brew &>/dev/null; then
```
MikeMcQuaid, line 5: I think it's worth leaving around this code, personally, to handle self-hosted Linux workers and in case they remove Homebrew on Linux from the Linux worker on Actions again.

setup-homebrew/main.sh
```diff
@@ -1,49 +1,48 @@
...
-        git checkout --force -B master FETCH_HEAD
-        cd -
-    ;;
-esac
-
-if [ "$RUNNER_OS" = "Linux" ]; then
+# brew
+if [[ "$GITHUB_REPOSITORY" =~ ^.+/brew$ ]]; then
+    rm -rf "$GITHUB_WORKSPACE"
+    ln -s "$HOMEBREW_REPOSITORY" "$GITHUB_WORKSPACE"
```
MikeMcQuaid, line 23:
```suggestion
    NEW_GITHUB_WORKSPACE="$HOMEBREW_REPOSITORY"
```

Can do that in each of the `if`/`elif`s and then have a `if [[ -n "$NEW_GITHUB_WORKSPACE" ]]` which does all of this (shared) logic with `ln -s "$NEW_GITHUB_WORKSPACE" "$GITHUB_WORKSPACE"`

dawidd6: I actually revisited it and those lines aren't always the same. I'd like to keep it the way it is now.

setup-homebrew/main.sh
dawidd6: Done

setup-homebrew/main.sh
MikeMcQuaid: Works for me 👍🏻

setup-homebrew/main.sh
```diff
@@ -22,28 +17,44 @@ else
...
+# core taps
+elif [[ "$GITHUB_REPOSITORY" =~ ^.+/(home|linux)brew-core$ ]]; then
+    rm -rf "$GITHUB_WORKSPACE"
+    ln -s "$HOMEBREW_CORE_REPOSITORY" "$GITHUB_WORKSPACE"
+    git fetch origin "$GITHUB_SHA"
+    git checkout --force -B master FETCH_HEAD
+# third-party taps
+elif [[ "$GITHUB_REPOSITORY" =~ ^.+/homebrew-.+$ ]]; then
+    mkdir -p "$(dirname "$HOMEBREW_TAP_REPOSITORY")"
+    ln -s "$GITHUB_WORKSPACE" "$HOMEBREW_TAP_REPOSITORY"
```
MikeMcQuaid, line 54: Might want to still delete this in case it's not around?

dawidd6: https://github.com/Homebrew/actions/commit/067bcc9c26756746fccc540638150334a38e220a

PR #30, opened 2020-07-31
label-pull-requests: add `missing_content` regex

SeekingMeaning:
Not sure if this is necessary or even works

Related: Homebrew/homebrew-core#58903

### DIFF
--- a/label-pull-requests/README.md
+++ b/label-pull-requests/README.md
@@ -20,6 +20,11 @@ An action that labels latest pull requests with given criteria.
               "label": "bottle unneeded",
               "content": "bottle :unneeded",
               "path": "Formula/.+"
+          },
+          {
+              "label": "missing license",
+              "missing": "license \"[^"]+\"",
+              "path": "Formula/.+"
           }
       ]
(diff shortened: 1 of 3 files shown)


jonchang:
Maybe `missing_content` instead of `missing` to be clearer?

jonchang:
I think we do want this, though, since it should be our goal to get a license into every formula and it will help to flag formulae during e.g. version bumps that they don't have a license.

dawidd6:
I'm wondering if having a field for matching PR title would be useful to us. For example if someone names their PR like: `Update xyz to a.b.c`, then we would slap a label `wrong title` or something.

dawidd6:
Also maybe we can do the unlabel thing too. When PR changes and a label doesn't apply anymore, then delete it?

MikeMcQuaid:
> Also maybe we can do the unlabel thing too. When PR changes and a label doesn't apply anymore, then delete it?

Yes, this would be good.

dawidd6:
Good job @SeekingMeaning !

PR #32, opened 2020-08-05
workflows/setup-homebrew: run test-bot.

MikeMcQuaid:
This is a better test because:

- it runs `brew config` for us
- this is what we're expecting users to run
- this is/shortly will be what `brew tap-new` is doing

### DIFF
--- a/setup-homebrew/README.md
+++ b/setup-homebrew/README.md
@@ -27,3 +27,13 @@ This also sets up the variables necessary to cache the gems installed by Homebre
   if: steps.cache.outputs.cache-hit != 'true'
   run: brew install-bundler-gems
 ```
+
+If you're using Linux, it will be quicker to use the GitHub Actions provided Ruby:
+
+```yaml
+- name: Set up Ruby
+  if: matrix.os == 'ubuntu-latest'
+  uses: actions/setup-ruby@master
+  with:
+    ruby-version: '2.6'
+```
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
Making some changes in test-bot to improve this: https://github.com/Homebrew/homebrew-test-bot/pull/460

setup-homebrew/README.md
```diff
@@ -27,3 +27,13 @@ This also sets up the variables necessary to cache the gems installed by Homebre
   if: steps.cache.outputs.cache-hit != 'true'
   run: brew install-bundler-gems
 ```
+
+If you're using Linux, it will be quicker to use the GitHub Actions provided Ruby:
```
dawidd6, line 31: Why is that? I thought we supply portable Ruby.

MikeMcQuaid: We do but it's way quicker to use the (already installed) version in GitHub Actions which requires only a symlink change.

dawidd6: You are saying that our portable Ruby is somehow slower than the one provided by GHA?

MikeMcQuaid: I'm saying that installing it is: yes.

dawidd6: But AFAIK portable Ruby is already installed with Homebrew in GHA VM images, so the cost of installing it on runtime is zero.

MikeMcQuaid: Last time I checked it wasn't but I could well be wrong here?

dawidd6: It is installed, there is no indication in logs that would tell it is being installed each time a workflow job is run and in `homebrew-core` we do not setup Ruby anywhere now.

dawidd6: See this test workflow run:
https://github.com/dawidd6/homebrew-test/runs/952878355?check_suite_focus=true

PR #53, opened 2020-08-09
setup-homebrew: don't chmod non-existing folder in docker container

iMichka:


dawidd6:
Maybe we can just replace /home/runner with $HOME?

MikeMcQuaid:
Thanks @iMichka!

PR #55, opened 2020-08-13
Various cleanup

MikeMcQuaid:
- add consistent spacing
- only use multiline `run` when necessary
- remove identical, implicit `name` usage
- use `actions/cache@v1` for compatibility

### DIFF
--- a/.github/workflows/brew-script.yml
+++ b/.github/workflows/brew-script.yml
@@ -16,6 +16,7 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+
       - name: Script
         uses: ./brew-script/
         with:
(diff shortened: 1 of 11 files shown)


maxim-belkin:
actions/cache@v1 has some limitations: [link](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)

MikeMcQuaid:
> actions/cache@v1 has some limitations: [link](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)

@maxim-belkin `v2` is actively broken for us.

PR #57, opened 2020-08-14
main.sh: allow to set core remote

iMichka:
We are transitioning from liuxbrew-core to homebrew-core,
so we need to be able to checkout the right repository.

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -12,7 +12,11 @@ if ! which brew &>/dev/null; then
     HOMEBREW_CORE_REPOSITORY="$HOMEBREW_REPOSITORY/Library/Taps/homebrew/homebrew-core"
     sudo mkdir -p "$HOMEBREW_CORE_REPOSITORY"
     sudo rm -rf "$HOMEBREW_CORE_REPOSITORY"
-    sudo git clone --depth=1 https://github.com/Homebrew/linuxbrew-core "$HOMEBREW_CORE_REPOSITORY"
+    if [ ! -z "$HOMEBREW_CORE_GIT_REMOTE" ]
+    then
+        HOMEBREW_CORE_GIT_REMOTE=https://github.com/Homebrew/homebrew-core
+    fi
+    sudo git clone --depth=1 "$HOMEBREW_CORE_GIT_REMOTE" "$HOMEBREW_CORE_REPOSITORY"
 
     cd "$HOMEBREW_PREFIX"
     sudo mkdir -p bin etc include lib opt sbin share var/homebrew/linked Cellar


iMichka:
I made the default "homebrew-core", which means that it will work out of the box in homebrew-core, and will need to define the env variable in linuxbrew-core.

Bo98:
Considering this is also used elsewhere than the core repositories (like Homebrew/brew and formulae.brew.sh generation), I think it’s safer for this to match the default in brew (linuxbrew-core) for now.

We could however make it smart and check if `GITHUB_REPOSITORY` is homebrew-core and adjust accordingly.

Bo98:
I think cloning does work properly, but `brew doctor` loudly complains if it’s not linuxbrew-core. Setting `HOMEBREW_CORE_GIT_REMOTE` in the core tap conditional should be sufficient.

dawidd6:
Or maybe just set HOMEBREW_FORCE_HOMEBREW_ON_LINUX in workflow file

iMichka:
For more context, this is related to this PR: https://github.com/Homebrew/homebrew-core/pull/59372
I would like to start to build some linux-only formulae in the homebrew-core repo. This is done within our Linux build container.

The current action checks out linuxbrew-core, but I want to have homebrew-core

Bo98:
> The current action checks out linuxbrew-core

It doesn't appear to:

```
'/__w/homebrew-core/homebrew-core' -> '/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-core'
From https://github.com/Homebrew/homebrew-core
 * branch                  cb1cc140bc8f8a6cc095dcd942b8a31d613a42d2 -> FETCH_HEAD
Reset branch 'master'
Your branch and 'origin/master' have diverged,
and have 22 and 36264 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
```

The `brew doctor` also indicates it's correctly checking out homebrew-core. The problem is `brew` explicitly rejects such setup.

For example in a fresh Docker container I can do:

```sh
cd $(brew --repo homebrew/core)
git remote set-url origin https://github.com/Homebrew/homebrew-core
brew doctor
```

and I'll get:

```
Warning: Suspicious https://github.com/Homebrew/linuxbrew-core git origin remote found.
The current git origin is:
  https://github.com/Homebrew/homebrew-core

With a non-standard origin, Homebrew won't update properly.
You can solve this by setting the origin remote:
  git -C "/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-core" remote set-url origin https://github.com/Homebrew/linuxbrew-core
```

which is an identical error as I see in the CI.

> Or maybe just set HOMEBREW_FORCE_HOMEBREW_ON_LINUX in workflow file

The problem is that this does more than just change git remote. It also changes the default prefix which we do not want:

```
$ HOMEBREW_FORCE_HOMEBREW_ON_LINUX=1 brew doctor

Warning: Your Homebrew's prefix is not /usr/local.
Some of Homebrew's bottles (binary packages) can only be used with the default
prefix (/usr/local).
You will encounter build failures with some formulae.
Please create pull requests instead of asking for help on Homebrew's GitHub,
Discourse, Twitter or IRC. You are responsible for resolving any issues you
experience while you are running this unsupported configuration.
```

while `HOMEBREW_CORE_GIT_REMOTE` can limit it to just that:

```
$ HOMEBREW_CORE_GIT_REMOTE=https://github.com/Homebrew/homebrew-core brew doctor
Your system is ready to brew.
```

(e.g. `echo "::set-env name=HOMEBREW_CORE_GIT_REMOTE::https://github.com/$GITHUB_REPOSITORY"` here or alternatively just set it in the workflow file)

iMichka:
Ok I used HOMEBREW_CORE_GIT_REMOTE in my workflow and the warning is gone. Thanks. This will do for now.

MikeMcQuaid:
I think this should be a configuration option for the Action which selects the repository based on user preference and otherwise it uses the `GITHUB_REPOSITORY` (as it does below).

PR #58, opened 2020-08-17
setup-homebrew: using composite

dawidd6:
https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action

This way, we call the script directly, rather than through NodeJS.


reitermarkus:
I think you need to explicitly set the outputs, like here:

https://github.com/Homebrew/homebrew-cask/blob/ae3969463036755deb88f0ca669d5980088e0f15/.github/actions/checkout_cask_pr/action.yml#L1-L7

setup-homebrew/action.yml
```diff
@@ -10,5 +10,9 @@ outputs:
   gem-hash:
     description: Homebrew's Ruby Gemfile.lock sha256sum
 runs:
-  using: node12
-  main: main.js
+  using: composite
+  steps:
+    - run: chmod +x $GITHUB_ACTION_PATH/main.sh
```
MikeMcQuaid, line 15: Will this show up in the output? Is it avoidable by doing `chmod +x` on the file in this repository? Alternatively, could do `bash $GITHUB_ACTION_PATH/main.sh` maybe?

dawidd6: I did `chmod +x main.sh` locally and committed it.

dawidd6:
@reitermarkus thanks. Turned out we had wrong output names in `action.yml` too.

PR #59, opened 2020-08-19
label-pull-requests: label if match, unlabel if no match

dawidd6:
This PR changes the Action's behavior to actually remove labels from PRs that no longer apply.

For example:
1. Contributor makes a bump PR for a formula
2. @BrewTestBot labels the PR with `deprecated license`
3. Contributor fixes the license
4. @BrewTestBot removes the label from PR

### DIFF
--- a/label-pull-requests/main.js
+++ b/label-pull-requests/main.js
@@ -27,7 +27,7 @@ async function main() {
             per_page: 50
         })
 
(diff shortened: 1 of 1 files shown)


dawidd6:
Looks like it works
https://github.com/Homebrew/homebrew-core/runs/1005149668?check_suite_focus=true

PR #60, opened 2020-08-19
brew-script: using composite

dawidd6:


brew-script/action.yml
```diff
@@ -13,5 +13,19 @@ inputs:
...
+      shell: bash
+    - run: brew ruby /tmp/brew-script-$GITHUB_SHA
+      shell: bash
+      env:
+        HOMEBREW_GITHUB_API_TOKEN: ${{inputs.token}}
+        HOMEBREW_NO_ENV_FILTERING: 1
+        HOMEBREW_NO_AUTO_UPDATE: 1
+        HOMEBREW_NO_ANALYTICS: 1
+        HOMEBREW_COLOR: 1
+        RUBYOPT: -W0
```
MikeMcQuaid, line 31: Why set this?

brew-script/action.yml
```diff
@@ -13,5 +13,19 @@ inputs:
...
+        ${{inputs.script}}
+        EOF
+      shell: bash
+    - run: brew ruby /tmp/brew-script-$GITHUB_SHA
+      shell: bash
+      env:
+        HOMEBREW_GITHUB_API_TOKEN: ${{inputs.token}}
+        HOMEBREW_NO_ENV_FILTERING: 1
+        HOMEBREW_NO_AUTO_UPDATE: 1
+        HOMEBREW_NO_ANALYTICS: 1
```
MikeMcQuaid, line 29: Why set this?

brew-script/action.yml
dawidd6: So one does not see messages like this:
```
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/language/python.rb:8: warning: Insecure world writable dir /home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/shims/scm in PATH, mode 040777

```

brew-script/action.yml
dawidd6: Could remove it if you think it isn't needed. Just thought we don't want analytics from this Action.

brew-script/action.yml
MikeMcQuaid: We should fix that message either in Homebrew or, ideally, in the environment we execute this in.

brew-script/action.yml
MikeMcQuaid: Yeh, we don't want to disable analytics anywhere really.

dawidd6: https://github.com/Homebrew/actions/commit/faee221d871037213f8d0e936fb113246b003539

brew-script/action.yml
dawidd6: https://github.com/Homebrew/brew/pull/8407

Deleted in: https://github.com/Homebrew/actions/commit/f4b98344cb593dc08583251f91d6701719a0c873

brew-script/action.yml
MikeMcQuaid: Thanks on both!

PR #89, opened 2020-09-15
label-pull-request: add `except`

SeekingMeaning:
This adds an `except` field to `label-pull-request` that allows files to be excluded by absolute path, i.e. without regex

Example:

```json
{
    "label": "legacy",
    "path": "Formula/.+@.+",
    "except": [
        "Formula/python@3.8",
        "Formula/python@3.9"
    ]
}
```

Related: Homebrew/homebrew-core#61163

### DIFF
--- a/label-pull-requests/main.js
+++ b/label-pull-requests/main.js
@@ -74,6 +74,7 @@ async function main() {
                 if (
                     (!constraint.status || file.status == constraint.status) &&
                     (!constraint.path || file.filename.match(constraint.path)) &&
+                    (!constraint.except || (Array.isArray(constraint.except) ? constraint.except.includes(file.filename) : file.filename == constraint.except)) &&
                     (!constraint.content || file.content.match(constraint.content)) &&
                     (!constraint.missing_content || !file.content.match(constraint.missing_content))
                 ) {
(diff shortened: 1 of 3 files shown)


label-pull-requests/main.js
```diff
@@ -74,6 +74,7 @@ async function main() {
                 if (
                     (!constraint.status || file.status == constraint.status) &&
                     (!constraint.path || file.filename.match(constraint.path)) &&
+                    (!constraint.except || (Array.isArray(constraint.except) ? constraint.except.includes(file.filename) : file.filename == constraint.except)) &&
```
MikeMcQuaid, line 77: Might be nice to split this onto multiple lines or a separate variable.

MikeMcQuaid (approved):

label-pull-requests/main.js
SeekingMeaning: What do you think about this?

```js
(!constraint.except ||
    (Array.isArray(constraint.except) ?
        constraint.except.includes(file.filename) :
        file.filename == constraint.except)) &&
```

MikeMcQuaid: Looks better! Think ternary would be nice to avoid here but don't feel strongly.

PR #90, opened 2020-09-22
Fix cache path.

reitermarkus:
Since we override `GITHUB_WORKSPACE`, the path we use for caching is wrong inside the `actions/cache@v2` action.

Problem:

The `cache` action will run inside of the resolved `GITHUB_WORKSPACE` path, i.e. in `HOMEBREW_REPOSITORY` or inside of a tap directory. The cache path is still resolved relative to the unresolved `GITHUB_WORKSPACE` variable.

Solution:

~~Set the `gems-path` output relative to the unresolved `GITHUB_WORKSPACE` instead of to an absolute path.~~

Reverse the link direction, so that `realpath` will be equal to `GITHUB_WORKSPACE`.

### DIFF
--- a/.github/workflows/setup-homebrew.yml
+++ b/.github/workflows/setup-homebrew.yml
@@ -13,8 +13,8 @@ jobs:
     strategy:
       matrix:
         os:
+          - macos-latest
           - ubuntu-latest
-          - macOS-latest
     runs-on: ${{matrix.os}}
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


MikeMcQuaid (approved):
Fine with me, good catch 👍🏻

reitermarkus:
This doesn't seem to work for `Homebrew/brew` because the `brew` symlink gets broken. We are not caching in `Homebrew/brew` anyways, so doing this only for taps is probably fine.

reitermarkus:
Also won't work for taps because `Tap.for_path` won't work …

Using this workaround for now: https://github.com/Homebrew/homebrew-cask/pull/89686

PR #95, opened 2020-09-29
workflows: use pull_request event trigger

dawidd6:
Using actions/checkout with pull_request_target results in wrong ref (master) being checked out when PR is opened.


MikeMcQuaid:
Nice catch.

PR #98, opened 2020-10-03
setup-homebrew: always set origin HEAD

jonchang:
Prevents errors such as in [this workflow run](https://github.com/dawidd6/homebrew-test-tap-new/runs/1198674370?check_suite_focus=true) from appearing.

```
brew pr-pull --debug --tap=$GITHUB_REPOSITORY $PULL_REQUEST
Error: undefined method `split' for nil:NilClass
Please report this issue:
https://docs.brew.sh/Troubleshooting
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/dev-cmd/pr-pull.rb:385:in `block in pr_pull'
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/dev-cmd/pr-pull.rb:378:in `each'
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/dev-cmd/pr-pull.rb:378:in `pr_pull'
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/brew.rb:119:in `<main>'
```

PR #99, opened 2020-10-03
setup-homebrew: explicitly fetch upstream refs

jonchang:
Alterate strategy from #98. Whenever we fetch something from a remote that's not HEAD (e.g. `git fetch origin $SHA`) also explicitly fetch HEAD.

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -66,7 +66,7 @@ if [[ "$GITHUB_REPOSITORY" =~ ^.+/brew$ ]]; then
         ln -vs "$HOMEBREW_REPOSITORY" "$GITHUB_WORKSPACE"
     fi
     git remote set-url origin "https://github.com/$GITHUB_REPOSITORY"
-    git_retry fetch --tags origin "$GITHUB_SHA"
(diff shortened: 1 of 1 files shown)


dawidd6:
Still nope:
https://github.com/dawidd6/homebrew-test-tap-new/runs/1202518962?check_suite_focus=true

jonchang:
OK, another instance where carefully reading the manpage was in order. From `git-fetch(1)`:

> When no `<refspec>`s appear on the command line, the refs to fetch are read from `remote.<repository>.fetch` variables instead.
> 
> Typically such a variable may look like this: `+refs/heads/*:refs/remotes/origin/*`

dawidd6:
Works now:
https://github.com/dawidd6/homebrew-test-tap-new/runs/1203349212?check_suite_focus=true

dawidd6 (approved):

jonchang:
Awesome, thanks for taking the time to debug this!

MikeMcQuaid:
Can you explain why/when this is needed? This looks like we're now fetching all upstream branches every time unnecessarily now.

PR #104, opened 2020-10-08
Allow alternative branch names to setup-homebrew

Rylan12:
Allow `setup-homebrew` to be run on taps that don't use `master` as the main branch name.

GitHub has switched its default branch name to `main` so new taps that want to use the `setup-homebrew` action may use this as their main branch name. The action currently checks out the `master` branch by default, so this PR adds an optional setting for a different branch name (still defaults to `master`).

This is my first time diving into GitHub actions. What's the best way to test changes?

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -142,7 +142,10 @@ else
         ln -vs "$HOMEBREW_TAP_REPOSITORY" "$GITHUB_WORKSPACE"
         git_retry fetch origin "$GITHUB_SHA" '+refs/heads/*:refs/remotes/origin/*'
         git remote set-head origin --auto
-        git checkout --force -B master FETCH_HEAD
+
+        head="$(git -C "$DIR" symbolic-ref refs/remotes/origin/HEAD)"
+        head="${head#refs/remotes/origin/}"
+        git -C "$DIR" checkout --force -B "$head" origin/HEAD
         cd -
     fi
 


jonchang:
You should be able to automatically determine the upstream branch name with `git symbolic-ref`. That was what #98 and #99 were intended to permit.

dawidd6:
Aren't we checking out FETCH_HEAD here and just creating a new local branch named "master" from it?

Rylan12:
I'm playing with a personal tap and the `publish` workflow (triggered by `pr-pull`) is not working as expected. [Here's the run I'm looking at](https://github.com/Rylan12/homebrew-personal/runs/1228534129?check_suite_focus=true).

What's happening I think is that `setup-homebrew` and `brew pr-pull` are run in the `master` branch but the `git-try-push` action is set to push to the `main` branch. First it checks out `main` and then pushes, but the changes made by `pr-pull` were made on the `master` branch so nothing get's pushed.

MikeMcQuaid:
> I agree, though, detecting the upstream branch automatically would be much better. I'm not familiar with `git symbolic-ref` but I'll look into it.

Agreed. 

```bash
    head="$(git -C "$DIR" symbolic-ref refs/remotes/origin/HEAD)"
    head="${head#refs/remotes/origin/}"
    git -C "$DIR" checkout --force -B "$head" origin/HEAD
```

This code's definition of `$head` is "whatever the upstream default branch is named" so could work here.

Rylan12:
I didn't know about either `git symbolic-ref` or the string removal with `#` until now, so thanks for exposing me!

MikeMcQuaid (approved):

MikeMcQuaid:
Nice work @Rylan12!

miccal:
We are seeing the following error for the CI tests in the `homebrew-cask` repo's:
```
/home/runner/work/_actions/Homebrew/actions/master/setup-homebrew/main.sh: line 146: DIR: unbound variable
```
Is this something that needs to be fixed here or in the Cask repo's?

Rylan12:
Ah, I think this should be `HOMEBREW_TAP_REPOSITORY` not `DIR`. Will open a follow up PR.

PR #109, opened 2020-10-09
setup-homebrew: fix branch detection

Rylan12:
Follow up to #104

As pointed out in https://github.com/Homebrew/actions/pull/104#issuecomment-706309912, the PR adding branch detection fails for non-core taps because it tried to reference the `DIR` variable which is never set.

I believe `DIR` should be changed to `HOMEBREW_TAP_REPOSITORY`.

Again, I'm not really sure how to go about testing this change.

CC: @miccal


miccal:
Thank you for the quick fix @Rylan12.

PR #110, opened 2020-10-11
git-try-push: exit from loop after successful pull

dawidd6:
https://github.com/Homebrew/homebrew-core/runs/1237640124?check_suite_focus=true

cc @jonchang


jonchang:
:+1: good catch!

PR #115, opened 2020-10-15
Revert "Allow alternative branch names to setup-homebrew"

dawidd6:
This commit breaks checking out pull request head on third-party taps.
Reverting it for now.

Example workflow run: https://github.com/dawidd6/homebrew-tap/pull/189/checks?check_run_id=1258197042

@Rylan12

PR #116, opened 2020-10-15
setup-homebrew: checkout to repo default branch

dawidd6:
Another take on:
https://github.com/Homebrew/actions/pull/115

cc @Rylan12


Rylan12 (approved):
Without having done any testing, this seems like it should work

Rylan12:
Thanks, @dawidd6!

This seems to be working properly now 👍

PR #119, opened 2020-11-18
check-commit-format: do not automerge for patchelf or binutils

iMichka:
These need to be carefully tested and manually merged.

### DIFF
--- a/check-commit-format/main.js
+++ b/check-commit-format/main.js
@@ -65,6 +65,12 @@ async function main() {
                 message = `${short_sha} modifies non-formulae files (maintainers must merge manually)`
                 break
             }
+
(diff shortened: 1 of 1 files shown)


iMichka:
Other linux formulae should be auto-mergeable once we have setup the workflow, so I just want to disable automerge for patchelf and binutils.

MikeMcQuaid:
This doesn't feel related to the commit message format? I think this should be part of the labelling and then a label used to prevent auto merge.

iMichka:
Right, I was mislead and I did not understood what this code was doing.

I think I need to add this to "triage.yml" in the homebrew-core repo.

PR #124, opened 2020-11-28
label-pull-requests: add keep_if_no_match option

### LINKED ISSUE #122: check-commit-style toggles automerge-skip label
I've noticed that the `check-commit-style` action doesn't add the `automerge-skip` label appropriately. Instead, it seems to toggle whether the label is applied or not. For example, opening a PR with an invalid commit will cause `automerge-skip` to be added. If you then force push a new commit that is still invalid, the `automerge-skip` label is removed. It should not be because the commit is still invalid.

I've noticed this a few times in homebrew/core, but I've opened a PR on a custom tap that illustrates the problem blocking any homebrew/core changes: https://github.com/Rylan12/homebrew-development/pull/5

I created the PR based on https://github.com/Rylan12/homebrew-development/commit/e4b004c040c6b2d71f4c0f30ea75462c3b0346ed and the `automerge-skip` label [was applied](https://github.com/Rylan12/homebrew-development/pull/5#event-4043895364).

Then, I forced pushed the branch to https://github.com/Rylan12/homebrew-development/commit/e640a2d4b0397493f82dd5a1e65a9e8c0e31cbdd and the `automerge-skip` label [was removed](https://github.com/Rylan12/homebrew-development/pull/5#event-4043900728).

Both commits modify two files so the label should still be present.

I haven't done any testing yet to determine where the bug is. If I have time later, I'll try to take a look but I figured I'd file this in case someone else has an idea.

Rylan12:
Resolves #122

This adds a configuration option to the `label-pull-requests` workflow that allows for certain conditions to be marked with the `keep_if_no_match` setting. When set, `label-pull-requests` will not attempt to remove the label if it already exists and no longer matches. This will be used in homebrew/core by the `automerge-skip` label. Currently, the configuration for that label is this:

```json
{
  "label": "automerge-skip",
  "path": "Formula/(patchelf|binutils).rb"
}
```

This meant that when `label-pull-requests` was run, the `automerge-skip` label was removed for all PRs that didn't modify either `Formula/patchelf.rb` or `Formula/binutils.rb`. After this PR, I'll change the configuration to:

```json
{
  "label": "automerge-skip",
  "path": "Formula/(patchelf|binutils).rb",
  "keep_if_no_match": true
}
```

Now, `automerge-skip` will still be added to PRs that modify `Formula/patchelf.rb` or `Formula/binutils.rb` but won't be removed from PRs that don't.

### DIFF
--- a/label-pull-requests/README.md
+++ b/label-pull-requests/README.md
@@ -35,6 +35,11 @@ Will remove labels from a pull request that no longer apply.
               "label": "missing license",
               "missing_content": "license \"[^"]+\"",
               "path": "Formula/.+"
+          },
+          {
+              "label": "automerge-skip",
+              "path": "Formula/(patchelf|binutils).rb",
+              "keep_if_no_match": true
           }
       ]
 ```
--- a/label-pull-requests/main.js
+++ b/label-pull-requests/main.js
@@ -70,6 +70,12 @@ async function main() {
                     labelExists = true
                 }
 
+                // Continue if the label exists and we aren't going to remove
+                // it regardless of whether it applies or not
+                if (labelExists && constraint.keep_if_no_match) {
+                    continue
+                }
+
                 // Check constraints
                 constraintApplies = doesConstraintApply(constraint, file)
 


Rylan12:
Reopening from a branch on the main repo instead of my fork.

PR #148, opened 2021-03-10
Call git_retry without sudo

shivammathur:
This PR fixes the setup-homebrew's main.sh script by calling `git_retry` function without `sudo`.
Functions cannot be called with `sudo`, since brew is pre-installed on runners, this code block must have never run before.

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -33,12 +33,12 @@ if ! which brew &>/dev/null; then
     HOMEBREW_PREFIX=/home/linuxbrew/.linuxbrew
     HOMEBREW_REPOSITORY="$HOMEBREW_PREFIX/Homebrew"
     sudo mkdir -p "$HOMEBREW_PREFIX"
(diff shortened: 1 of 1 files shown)


dawidd6:
Won't we get denied permission while trying to clone after this change?

shivammathur:
@dawidd6 in that case, `sudo` can be used inside `git_retry`.

dawidd6:
It can, but shouldn't. This way git would be called as root every time in this Action. We don't want that. I'm actually in favor of deleting this whole chunk of code for good. Homebrew is always installed on our runners or the container that we use and this Action is mostly for our internal use + for people to use in their own taps.

MikeMcQuaid:
@dawidd6 Kill it. We needed to keep it around when there were macOS and Linux inconsistencies and when Linuxbrew was getting added and removed from the images but that's no longer the case.

shivammathur:
In that case, I will close this.

PR #149, opened 2021-03-12
setup-homebrew: remove Homebrew bootstrapping

dawidd6:
See the discussion here: https://github.com/Homebrew/actions/pull/148

In conclusion: this code is not used anymore nor is it tested on CI.

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -28,30 +28,10 @@ function git_retry {
   return 0
 }
 
-# Clone Homebrew/brew and Homebrew/linuxbrew-core if necessary.
-if ! which brew &>/dev/null; then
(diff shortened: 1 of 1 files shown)


setup-homebrew/main.sh
```diff
@@ -28,30 +28,10 @@ function git_retry {
   return 0
 }
 
-# Clone Homebrew/brew and Homebrew/linuxbrew-core if necessary.
-if ! which brew &>/dev/null; then
```
MikeMcQuaid, line 32: Could consider keeping this check around and exiting if it fails?

MikeMcQuaid (approved):

setup-homebrew/main.sh
dawidd6: Done

MikeMcQuaid (approved):

PR #153, opened 2021-03-19
Create action to setup commit signing

nandahkrishna:
This PR creates a new action to setup commit signing, to be used mainly with BrewTestBot. I thought creating a separate action would be useful in debugging issues with commit signing, and thus better than adding this to `git-user-config`.

Note: The test workflow I've added will not work unless the necessary secrets (`BREWTESTBOT_GPG_SIGNING_SUBKEY` and `BREWTESTBOT_GPG_SIGNING_SUBKEY_PASSPHRASE`) are accessible in this repository (cc @MikeMcQuaid).

Also, I was wondering if pushing a test signed commit to a branch to ensure that the "Verified" badge appears would be okay. Checking the output of `git log --show-signature -1` is not sufficient to ensure this (I noticed some issues when I was testing this functionality earlier).

### DIFF
--- a/.github/workflows/setup-commit-signing.yml
+++ b/.github/workflows/setup-commit-signing.yml
@@ -1,0 +1,49 @@
+name: Setup GPG commit signing
+
+on:
+  push:
+    paths:
+      - '**setup-commit-signing**'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'node_modules/**'
+
+jobs:
+  commit-signing:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@main
+
+      - name: Configure git user
+        uses: ./git-user-config/
+        with:
+          username: BrewTestBot
+
+      - name: Set up commit signing
+        id: set-up-commit-signing
+        uses: ./setup-commit-signing/
+        with:
+          signing_key: ${{ secrets.BREWTESTBOT_GPG_SIGNING_SUBKEY }}
+
+      - name: Make changes and commit them
+        run: |
+          git checkout -b test-setup-commit-signing
+
+          touch test.txt
+          echo "Hello, Homebrew!" >> test.txt
+
+          git add test.txt
+          git commit -m "test.txt: create and add content."
+        env:
+          GPG_PASSPHRASE: ${{ secrets.BREWTESTBOT_GPG_SIGNING_SUBKEY_PASSPHRASE }}
+
+      - name: Test
+        run: |
+          git log --show-signature -1 | grep "gpg: Good"
+
+      - name: Push commits
+        uses: ./git-try-push/
+        with:
+          token: ${{ github.token }}
+          branch: test-setup-commit-signing
--- a/setup-commit-signing/action.yml
+++ b/setup-commit-signing/action.yml
@@ -1,0 +1,17 @@
+name: setup-commit-signing
+description: Set up GPG commit signing
+author: nandahkrishna
+branding:
+  icon: user-check
+  color: green
+inputs:
+  signing_key:
+    description: Secret key to use for signing commits
+    required: true
+runs:
+  using: composite
+  steps:
+    - run: |
+        "$GITHUB_ACTION_PATH/main.sh" '${{ inputs.signing_key }}'
+      shell: bash
+      id: setup
--- a/setup-commit-signing/main.sh
+++ b/setup-commit-signing/main.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -euo pipefail
+
+GPG_SIGNING_KEY="${1}"
+
+export GNUPGHOME="$HOME/.gnupg/"
+mkdir -p "$GNUPGHOME"
+chmod 0700 "$GNUPGHOME"
+
+GPG_EXEC=$(command -v gpg)
+
+# Wrapper script to use passphrase non-interactively with git
+GPG_WITH_PASSPHRASE=$(mktemp)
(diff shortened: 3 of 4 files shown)


MikeMcQuaid:
> Also, I was wondering if pushing a test signed commit to a branch to ensure that the "Verified" badge appears would be okay. 

Makes sense 👍🏻

setup-commit-signing/action.yml
```diff
@@ -0,0 +1,21 @@
...
+branding:
+  icon: user-check
+  color: green
+inputs:
+  signing_key:
+    description: Secret key to use for signing commits
+    required: true
+  passphrase:
+    description: Secret key passphrase
+    required: false
```
MikeMcQuaid, line 13: How does this work if it's not present (or is it just that secret keys technically don't require passphrases)?

setup-commit-signing/main.sh
```diff
@@ -0,0 +1,25 @@
...
+set -euo pipefail
+
+SIGNING_KEY="${1}"
+PASSPHRASE="${2}"
+
+export GNUPGHOME="$HOME/.gnupg/"
+mkdir -p "$GNUPGHOME"
+chmod 0700 "$GNUPGHOME"
+
+git config --global gpg.program /usr/bin/gpg
```
MikeMcQuaid, line 12: Might want to check this exists and error otherwise. Also might want to use one from the `PATH` for macOS.

setup-commit-signing/main.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Wrapper script to use passphrase non-interactively with git
+if [ -z "$PASSPHRASE" ]; then
+  echo '/usr/bin/gpg --pinentry-mode loopback --passphrase "$PASSPHRASE" --batch --no-tty "$@"' > /tmp/gpg-with-passphrase
+  chmod +x /tmp/gpg-with-passphrase
```
MikeMcQuaid, line 17: Putting this in `/tmp` seems a little insecure?

MikeMcQuaid:
Is this failing because the key needs added to this repository?

nandahkrishna:
> Is this failing because the key needs added to this repository?

Yes, the key and its passphrase.

setup-commit-signing/action.yml
nandahkrishna: Keys don't require a passphrase, although it is recommended to have one.

MikeMcQuaid: I think we should probably require it in this workflow because not setting one seems like A Very Bad Idea.

.github/workflows/setup-commit-signing.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: ./git-user-config/
+        with:
+          username: BrewTestBot
+
+      - name: Set up commit signing
+        id: set-up-commit-signing
+        uses: ./setup-commit-signing/
+        with:
+          signing_key: ${{ secrets.BREWTESTBOT_GPG_SIGNING_SUBKEY }}
+          passphrase: ${{ secrets.BREWTESTBOT_GPG_SIGNING_SUBKEY_PASSPHRASE }}
```
MikeMcQuaid, line 28: These are only going to be set on `push`, not `pull_request`.

Rylan12:
I just want to clarify, the `BREWTESTBOT_GPG_SIGNING_SUBKEY` and `BREWTESTBOT_GPG_SIGNING_SUBKEY_PASSPHRASE` secrets need to be added on a per-repo basis to be used. So the only reason it needs to be added here is so the test passes, not because each execution uses the secrets from this repository. So someone in another repo who uses this action will need to add their own secrets and won't be able to sign commits using @BrewTestBot's GPG key, right?

nandahkrishna:
Silly enough, I didn't realise that the workflow does not need the passphrase for the secret key as an input. We need the passphrase only when committing something (which calls the temporary wrapper script).

The passphrase is used from an environment variable (`PASSPHRASE`) by the script, so we need to make sure it's set with `env`.

nandahkrishna:
> I just want to clarify, the `BREWTESTBOT_GPG_SIGNING_SUBKEY` and `BREWTESTBOT_GPG_SIGNING_SUBKEY_PASSPHRASE` secrets need to be added on a per-repo basis to be used.

Yes, and I believe it's possible to restrict organisation secrets to some repositories ([docs](https://docs.github.com/en/actions/reference/encrypted-secrets#about-encrypted-secrets)). It's necessary to have access to these secrets here for the tests alone.

> So someone in another repo who uses this action will need to add their own secrets and won't be able to sign commits using @BrewTestBot's GPG key, right?

Yes. The action merely takes a secret key and imports it, and does the necessary `git config`.

.github/workflows/setup-commit-signing.yml
```diff
@@ -0,0 +1,49 @@
...
+        run: |
+          git checkout -b test-setup-commit-signing
+
+          touch test.txt
+          echo "Hello, Homebrew!" >> test.txt
+
+          git add test.txt
+          git commit -m "test.txt: create and add content."
+        env:
+          PASSPHRASE: ${{ secrets.BREWTESTBOT_GPG_SIGNING_SUBKEY_PASSPHRASE }}
```
MikeMcQuaid, line 39:
```suggestion
          GPG_PASSPHRASE: ${{ secrets.BREWTESTBOT_GPG_SIGNING_SUBKEY_PASSPHRASE }}
```

maybe?

dawidd6:
How about creating a dummy GPG key just for testing and paste it directly into the input? No need for secrets here IMO.

nandahkrishna:
We could do that. It would require adding the public key to BrewTestBot's account using the GitHub API, and then removing it at the end of the workflow during cleanup.

dawidd6:
Which would require a GitHub token to do so heh. Nah let it be like it is now then.

MikeMcQuaid (approved):
Great work!

PR #154, opened 2021-03-23
setup-homebrew: allow to run as normal user in container

dawidd6:
By modifying permissions of some directories in Docker container we are able to run `brew` as `linuxbrew` user inside it without any friction (at least looking at `coreutils` testing PR).

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -43,7 +43,7 @@ HOMEBREW_CASK_REPOSITORY="$HOMEBREW_REPOSITORY/Library/Taps/homebrew/homebrew-ca
 # Do in container or on the runner
 if grep -q actions_job /proc/1/cgroup; then
     # Fix permissions to give normal user access
-    sudo chown -R "$(whoami)" "$HOME" "$GITHUB_WORKSPACE/.."
+    sudo chown -R "$(whoami)" "$HOME" "$PWD/.."
 else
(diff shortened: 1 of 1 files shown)


setup-homebrew/main.sh
```diff
@@ -40,8 +40,14 @@ HOMEBREW_REPOSITORY="$(brew --repo)"
...
 
-# Add brew to PATH
-echo "$HOMEBREW_PREFIX/bin" >> $GITHUB_PATH
+# Do in container or on the runner
+if grep -q actions_job /proc/1/cgroup; then
+    # Fix permissions to give normal user access
+    sudo chown -R "$(whoami)" "$HOME" "$PWD/.."
+else
+    # Add brew to PATH
+    echo "$HOMEBREW_PREFIX/bin" >> $GITHUB_PATH
```
sjackman, line 49: Is `$HOMEBREW_PREFIX/bin` already in `PATH` when running as a non-root user?

sjackman: Perhaps a comment explaining why this line isn't need when running as a non-root user?

dawidd6: It's already in PATH because it it our Docker container.

dawidd6: I'm gonna merge this now, feel free to suggest any comments afterwards.

PR #160, opened 2021-04-08
Make version format comparison check more strict.

reitermarkus:
Only allow decimal number changes (of any length) and hexadezimal changes (which match the length of the previous version).

### DIFF
--- a/.github/workflows/review-cask-pr.yml
+++ b/.github/workflows/review-cask-pr.yml
@@ -1,0 +1,26 @@
+name: Review Cask PR Unit Tests
+
+on:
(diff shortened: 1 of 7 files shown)


MikeMcQuaid (approved):
Looks good. Could this get some tests for some of the bad cases we've seen?

MikeMcQuaid:
> Could this get some tests for some of the bad cases we've seen?

Oh, sorry, missed that this is in Homebrew/actions so no worries if that's difficult to handle here.

reitermarkus:
Added some tests.

PR #170, opened 2021-05-06
wait-for-idle-runner (new action)

iMichka:
This new actions is needed for https://github.com/Homebrew/homebrew-core/pull/75951

### DIFF
--- a/.github/workflows/test-wait-for-idle-runner.yml
+++ b/.github/workflows/test-wait-for-idle-runner.yml
@@ -1,0 +1,23 @@
+name: Test create gcloud install
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    paths: 
+      - 'wait-for-idle-runner/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'node_modules/**'
+jobs:
+  syntax:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm install @octokit/action
+      - name: Check scripts
+        run: |
+          node --check wait-for-idle-runner/main.js
--- a/wait-for-idle-runner/Readme.md
+++ b/wait-for-idle-runner/Readme.md
@@ -1,0 +1,17 @@
+# Wait for idle runner GitHub Action
+
+An action that waits for an idle runner.
+The `runner-found` output is set to `true` or `false` based on whether or not the runner was found.
+
+If the runner was found, the `runner-idle` output is set to `true` or `false` based on whether the runner is idle.
+
+## Usage
+
+```yaml
+- name: Wait for idle runner
+  uses: Homebrew/actions/wait-for-idle-runner@master
+  with:
+    github_token: ${{ secrets.HOMEBREW_GITHUB_API_TOKEN }}
+    runner_name: linux-self-hosted-1
+    repo_name: ${{ github.event.repository.full_name }}
+```
--- a/wait-for-idle-runner/action.yml
+++ b/wait-for-idle-runner/action.yml
@@ -1,0 +1,26 @@
+name: Wait for idle runner
+description: Wait until a self-hosted runner is idle
+author: iMichka
+branding:
+  icon: message-square
+  color: purple
+inputs:
+  github_token:
+    description: GitHub token
+    required: true
+  runner_name:
+    description: Name of the self-hosted runner
+    required: true
+  repository_name:
+    description: Full name of the repository
+    required: true
+    default: ${{ github.event.repository.full_name }}
+outputs:
+  runner-found:
+    description: Runner exists (true/false)
+  runner-idle:
+    description: Runner is idle (true/false)
+runs:
+  using: node12
+  main: main.js
+  id: killable
--- a/wait-for-idle-runner/main.js
+++ b/wait-for-idle-runner/main.js
@@ -1,0 +1,35 @@
+const { Octokit } = require("@octokit/action");
+
+const octokit = new Octokit();
+
+const NUMBER_OF_ATTEMPTS = 3000;
+const TIME_BETWEEN_ATTEMPTS_SECONDS = 5000;
+
+async function main() {
+  for (var i=0; i < NUMBER_OF_ATTEMPTS; i++) {
+    const result = await octokit.request(`GET /repos/${process.env.REPOSITORY_NAME}/actions/runners`)
+
+    // Select the runner based on its name
+    runner = result.data["runners"].filter(function(runner) {
+      return runner.name == process.env.RUNNER_NAME;
+    })[0]
+
+    if (typeof runner == "undefined") {
+      console.log("::set-output name=runner-found::false");
+      return;
+    }
+
+    if (runner.busy == false) {
+      console.log("::set-output name=runner-found::true");
+      console.log("::set-output name=runner-idle::true");
+      return;
+    }
+
+    await new Promise(resolve => setTimeout(resolve, TIME_BETWEEN_ATTEMPTS_SECONDS));
+  }
+
+  console.log("::set-output name=runner-found::true");
+  console.log("::set-output name=runner-idle::false");
+}
+
+main()
--- a/wait-for-idle-runner/package.json
+++ b/wait-for-idle-runner/package.json
@@ -1,0 +1,4 @@
+{
+  "name": "wait-for-idle-runner",
+  "main": "main.js"
+}


iMichka:
I saw that there are some tests in the .github folder, but I'm a little bit unsure what I am supposed to add as test for this action.

wait-for-idle-runner/Readme.md
```diff
@@ -0,0 +1,17 @@
...
+
+```yaml
+- name: Wait for idle runner
+  uses: Homebrew/actions/wait-for-idle-runner@master
+  with:
+    github_token: ${{ secrets.HOMEBREW_GITHUB_API_TOKEN }}
+    runner_name: linux-self-hosted-1
+    repo_name: ${{ github.event.repository.full_name }}
+```
+
```
MikeMcQuaid, line 16:
```suggestion
```

wait-for-idle-runner/action.yml
```diff
@@ -0,0 +1,21 @@
...
+  color: purple
+inputs:
+  github_token:
+    description: GitHub token
+    required: true
+  runner_name:
+    description: Name of the self-hosted runner
+    required: true
+  repo_name:
+    description: Full name of the repo
```
MikeMcQuaid, line 15:
```suggestion
    description: Full name of the repository
```

wait-for-idle-runner/action.yml
```diff
@@ -0,0 +1,21 @@
...
+  icon: message-square
+  color: purple
+inputs:
+  github_token:
+    description: GitHub token
+    required: true
+  runner_name:
+    description: Name of the self-hosted runner
+    required: true
+  repo_name:
```
MikeMcQuaid, line 14:
```suggestion
  repository_name:
```

wait-for-idle-runner/main.js
```diff
@@ -0,0 +1,27 @@
+const { Octokit } = require("@octokit/action");
```
MikeMcQuaid, line 1: Is there any JS linting we could do of this file?  If we can't run it it'd be good to have some sort of linting.

iMichka: I have no clue about js linting. I can ask in the maintainer channel.

Note that no other js script in this repo seems to have .js linting.

I am willing to do the extra step to add js linting to this repo to get this merged, but on the other side it feels like yet another bigger project, especially if I have to fix js style issues in 10 other actions: js is not my programming language at all so this might take some more time.

MikeMcQuaid: > Note that no other js script in this repo seems to have .js linting.

That's because they are all actually run. As-is there's no CI here to provide guardrails to avoid breaking this script.

Rylan12: I'm not really the right person to ask about javascript stuff but I do know that you can use `node --check` to check for syntax errors which is definitely better than nothing.

```console
$ cat test.js
console.log("hello, world!"

$ node --check test.js
/Users/rylanpolster/Development/homebrew/actions/wait-for-idle-runner/test.js:1
console.log("hello, world!"
            ^^^^^^^^^^^^^^^

SyntaxError: missing ) after argument list
    at Object.compileFunction (node:vm:355:18)
    at wrapSafe (node:internal/modules/cjs/loader:1039:15)
    at checkSyntax (node:internal/main/check_syntax:66:3)
    at node:internal/main/check_syntax:39:3
```

iMichka: If the language was Python or Java I would have mocked the API calls and/or created a --dry-run version, but as this is javascript I am not sure if I want to spend more time on this right now. I added the `node --check` test.

wait-for-idle-runner/main.js
```diff
@@ -0,0 +1,27 @@
...
+const TIME_BETWEEN_ATTEMPTS_SECONDS = 5000;
+
+async function main() {
+  for (var i=0; i < NUMBER_OF_ATTEMPTS; i++) {
+    const result = await octokit.request(`GET /repos/${process.env.REPOSITORY_NAME}/actions/runners`)
+
+    // Select the runner based on it's name
+    runner = result.data["runners"].filter(function(runner) {
+      return runner.name == process.env.RUNNER_NAME;
+    })[0]
```
Rylan12, line 15: What happens if the runner name isn't found?

iMichka: I added a new check for this. The runner variable will be undefined, and in that case I am using an early return now.

.github/workflows/test-wait-for-idle-runner.yml
```diff
@@ -0,0 +1,23 @@
...
+      - 'package-lock.json'
+      - 'node_modules/**'
+jobs:
+  syntax:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
```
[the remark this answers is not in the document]
iMichka: I'm following the examples in upstream's readme: https://github.com/octokit/action.js
Don't know how to check if it works with another version.

wait-for-idle-runner/action.yml
```diff
@@ -0,0 +1,21 @@
...
+    description: GitHub token
+    required: true
+  runner_name:
+    description: Name of the self-hosted runner
+    required: true
+  repository_name:
+    description: Full name of the repository
+    required: true
+    default: ${{ github.event.repository.full_name }}
+runs:
```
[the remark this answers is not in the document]
iMichka: Fixed, you are right, the output was missing.

Not sure how to handle multiple outputs. The javascript code writes to the console, and if I want to set 2 outputs, I'm unsure if I can just log both?

I think for a first version, the current implementation is probably enough. I do not want to do premature optimisation here: I want to ship this and use it, and see if there are edge cases that we forgot. Changing the output can easily be done later on if necessary.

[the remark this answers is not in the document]
Rylan12: I'm pretty sure you just need to change the name that follows `name=` and precedes the `::value` in the log.

```js
console.log("::set-output name=runner-found::true"); // Sets `runner-found` output
console.log("::set-output name=runner-idle::false"); // Sets `runner-idle` output
```

I'm fine with merging as-is, though, if that's what you think is best.

wait-for-idle-runner/Readme.md
```diff
@@ -0,0 +1,15 @@
+# Wait for idle runner GitHub Action
+
+An action that waits for an idle runner.
+It returns "::set-output name=result::true" when the runner is idle.
```
Rylan12, line 4: I'd maybe say something like this instead, as it doesn't really _return_ that text

```suggestion
The `result` output is set to reflect the state of the runner.
```

Feel free to change that to better match what the `result` output really is.

wait-for-idle-runner/main.js
```diff
@@ -0,0 +1,33 @@
...
+
+    // Select the runner based on its name
+    runner = result.data["runners"].filter(function(runner) {
+      return runner.name == process.env.RUNNER_NAME;
+    })[0]
+
+    if (typeof runner == "undefined") {
+      // Return if the runner does not exist
+      console.log("::set-output name=result::Runner does not exist");
+      break;
```
Rylan12, line 20: Can we just `return` at this point?

Then, it might be a good idea to put a final `console.log` at the end of the function that sets the output to `Timed out` or something like that.

Rylan12 (commented):
Looks pretty good. A few more minor suggestions

MikeMcQuaid (commented):
👍🏻 once feedback addressed

iMichka:
I updated the outputs as asked.

MikeMcQuaid (approved):

wait-for-idle-runner/main.js
```diff
@@ -0,0 +1,35 @@
...
+  for (var i=0; i < NUMBER_OF_ATTEMPTS; i++) {
+    const result = await octokit.request(`GET /repos/${process.env.REPOSITORY_NAME}/actions/runners`)
+
+    // Select the runner based on its name
+    runner = result.data["runners"].filter(function(runner) {
+      return runner.name == process.env.RUNNER_NAME;
+    })[0]
+
+    if (typeof runner == "undefined") {
+      console.log("::set-output name=runner-found::false");
```
Rylan12, line 18: I don't think any changes are needed but I'm just curious if there's a default for `runner-idle` in this case. I think (although I could be wrong) that it defaults to some type of empty value that's treated as false?

It probably doesn't matter because the `runner-found` value should probably be checked before even bothering with `runner-idle`

wait-for-idle-runner/Readme.md
```diff
@@ -0,0 +1,15 @@
+# Wait for idle runner GitHub Action
+
+An action that waits for an idle runner.
+The `result` output is set to reflect the state of the runner.
```
Rylan12, line 4: Sorry to iterate more on this but I think something like this makes a little more sense now that `runner-found` and `runner-idle` are being used:

```suggestion
The `runner-found` output is set to `true` or `false` based on whether or not the runner was found.
If the runner was found, the `runner-idle` output is set to `true` or `false` based on whether the runner is idle.
```

Rylan12 (approved):
Looks good!

One more minor documentation update but other than that: 👍

wait-for-idle-runner/main.js
iMichka: `runner-found` needs to be checked first, so I did not add any `runner-idle` for this case. Not sure what happens with the value but it is probably set to nothing. I would not worry for this case too much right now, and we can improve is this is needed. The `runner-found` check is an edge case check I added because it was asked for in one of the reviews: initially I did not even plan for the runner to not exist, as it should always be there.

PR #191, opened 2021-05-23
Create composite action for Dispatch Build Bottle build job

GauthamGoli:
https://github.com/Homebrew/brew/issues/11101

### DIFF
--- a/dispatch-build-bottle/action.yml
+++ b/dispatch-build-bottle/action.yml
@@ -1,0 +1,124 @@
+name: Dispatch build bottle
+inputs:
+  formula:
+    description: "Formula name"
+    required: true
+  issue:
+    description: "Issue number, where comment on failure would be posted"
+    required: true
+  ignore-errors:
+    description: "Ignore errors (default: false)"
+    required: false
+  macos:
+    description: "macOS Version"
+    required: false
+  name:
+    description: "Name"
+    required: false
+  sender:
+    description: "Sender"
+    required: true
+  upload:
+    description: "Whether to upload built bottles or not (default: false)"
+    required: false
+runs:
+  using: "composite"
+  steps:
+    - name: ${{ inputs.formula }}
+      id: print_details
+      run: |
+        echo formula=${{ inputs.formula }}
+        echo sender=${{ inputs.sender }}
+        echo name=${{ inputs.name }}
+        echo ignore-errors=${{ inputs.ignore_errors }}
+        echo version=${{ inputs.macos }}
+        echo issue=${{ inputs.issue }}
+        echo upload=${{ inputs.upload }}
+        echo "::set-output name=repo::$(brew --repo ${{github.repository}})"
+
+    - name: Set up Homebrew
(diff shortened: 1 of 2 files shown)


dispatch-build-bottle/action.yml
```diff
@@ -0,0 +1,124 @@
...
+        echo "::set-output name=repo::$(brew --repo ${{github.repository}})"
+
+    - name: Set up Homebrew
+      id: set-up-homebrew
+      uses: Homebrew/actions/setup-homebrew@master
+
+    - name: Run brew test-bot --only-cleanup-before
+      if: github.repository == 'Homebrew/homebrew-core'
+      id: test-bot-only-cleanup-before
+      run: brew test-bot --only-cleanup-before
```
GauthamGoli, line 46: Steps which have  `if: github.repository == 'Homebrew/homebrew-core'`, were present only in `homebrew-core`'s `dispatch-build-bottle` action,  but not for `linuxbrew-core` 

Can this condition be modified to `if: always()` ?  I'm not sure and need input here

dispatch-build-bottle/action.yml
```diff
@@ -0,0 +1,124 @@
...
+      if: always()
+      env:
+        HOMEBREW_GITHUB_API_TOKEN: ${{secrets.GITHUB_TOKEN}}
+      run: |
+        mkdir ~/bottles
+        cd ~/bottles
+        brew test-bot \
+          --only-formulae \
+          --keep-old \
+          --skip-recursive-dependents \
```
GauthamGoli, line 63: `--skip-recursive-dependents` is not present in `homebrew-core`'s action's `step` ,
Does this step need to be separated  based on `github.repository` ?

dawidd6:
Did you test this Action? Cause I can tell right now that it might not work in current form. AFAIK composite Actions does not support running other Actions using `uses` keyword. Did something change recently around in this matter?

dawidd6:
```
Error: GauthamGoli/actions/dispatch-build-bottle-composite/dispatch-build-bottle/action.yml (Line: 27, Col: 7): Required property is missing: shell
Error: GauthamGoli/actions/dispatch-build-bottle-composite/dispatch-build-bottle/action.yml (Line: 41, Col: 7): Unexpected value 'uses'
...
```

Idea is good, but I don't think it will work (for now).

dispatch-build-bottle/action.yml
RandomDSdevel: &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;No, some users might (well?) then run into issues like the ones I and @reitermarkus fixed in Homebrew/brew#9457, Homebrew/homebrew-core#65917, and Homebrew/homebrew-cask#89714.  You can still loosen this condition to only run actions if they're triggered from within the Homebrew organization, though.

GauthamGoli:
As noted by @dawidd6  this will not work for now. More details are posted in the above linked parent issue on why this will not work.

PR #211, opened 2021-08-12
setup.sh: set and unset pipefail

iMichka:
Instead of removing completely the `pipefail` line for the whole script, do it only for the line which got broken by it.


iMichka:
@MikeMcQuaid I hope I got it right.

MikeMcQuaid (approved):
Perfect, thanks @iMichka 🎉 ❤️

PR #217, opened 2021-08-30
create-gcloud-instance: create machine with 4 vCPUs instead of 2

iMichka:
Based on gcloud's recommendation.
This should improve build times in homebrew-core


SMillerDev (approved):

iMichka:
I regenerated the runner as it was idle, it has now 4 vCPUs.

cho-m:
Maybe we can try finishing up one of the open PRs that use self-hosted and re-run.
- `boost boost-python` - **467m** - needs a rebase and some patches before re-run
- `swift` - **653m** - guaranteed to be slower than macOS since we use existing libraries there. Would be a good comparison PR, but it requires fixing LLVM linkage failure.
- `libffi` - **1433m** - it is currently faster than macOS, but could be state of dependents. Not sure on state of PR and would require blocking over 1 day of CI time.

Bo98:
Going to run Swift again soon so that'll be a good test.

Bo98:
Can compare Swift here: https://github.com/Homebrew/homebrew-core/pull/73984

It's actually not quite as fair of a comparison as it used to be. The Linux build now does build a bit more than macOS - it builds the stdlib as well as Foundation, libdispatch and XCTest ports for Linux.

With that said, it is definitely quicker than GitHub hosted runners. It also manages to keep up with macOS despite having less cores. (Our macOS VMs are hexacore machines with 15GB of RAM, but have the overhead of being a VM.)

carlocab:
Mojave is super fast; we should do that on all the other runners

Bo98:
Mojave is fundamentally a faster OS in its design - it's why I've stuck with it on my laptop (though I will have to make the jump to macOS 12 when it arrives). Stuff like the new system partitioning introduced some pretty serious regressions. Some of them were fixed in Big Sur.

In terms of CI, the actual hardware is identical. In fact, we now have "IO boost" enabled on Catalina and Big Sur (it doesn't work properly with Mojave for some reason even though it's meant to) and they are _still_ slower than Mojave. We'll definitely notice our CI times overall increasing when Mojave is dropped.

Bo98:
(Just realised you were maybe joking a bit in reference to Swift, but you did make a valid point - Mojave has always been faster than other OS versions since day 1.)

carlocab:
For context:

Screenshot 2021-08-31 at 23 01 06

PR #227, opened 2021-09-28
triage-issues.yml: update lock-threads parameter names

nandahkrishna:
The input parameter names have changed for the lock threads action (updated in #226). I've updated them based on https://github.com/Homebrew/.github/pull/37#issuecomment-929210701.


Bo98:
Actually this should go in Homebrew/.github.

Bo98:
Addressed upstream.

PR #244, opened 2021-12-08
setup-homebrew: apply self-hosted logic to Linux

Bo98:
(Re-)introduce `GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED` as we need to apply the self-hosted logic to Linux too.


dawidd6:
No change in brew or test-bot?

Bo98:
There will be. Just not gone through it all yet.

MikeMcQuaid (approved):

dawidd6 (approved):

PR #250, opened 2022-01-11
bump-formulae: use `brew bump`.

MikeMcQuaid:
Requires https://github.com/Homebrew/brew/pull/12702
Fixes https://github.com/Homebrew/brew/issues/5725

Use `brew bump` directly instead of the Ruby code here.

CC @dawidd6 @samford @chenrui333 @carlocab for thoughts and feedback


chenrui333 (approved):

carlocab (commented):
The Ruby script seems to handle installing `pipgrip` before calling `bump-formula-pr`, which isn't done in Homebrew/brew#12702.

If that handling is still necessary, we probably want to do that elsewhere (in `bump-formula-pr`, maybe?) before doing this.

MikeMcQuaid:
> The Ruby script seems to handle installing `pipgrip` before calling `bump-formula-pr`, which isn't done in [Homebrew/brew#12702](https://github.com/Homebrew/brew/pull/12702).

Agreed 👍🏻

chenrui333:
Nice catch, I realized I never used `bump` command myself in the past 😆

MikeMcQuaid:
@carlocab Good call, doing that in https://github.com/Homebrew/brew/pull/12703.

samford (approved):
Outside of the `pipgrip` issue (already addressed), there are some changes that will be needed in the `brew bump` PR to maintain this action's behavior (I added related suggestions/comments). I imagine we won't merge this PR until sometime after the `brew bump` one, so the changes here look fine to me 👍

MikeMcQuaid:
Thanks all!

PR #252, opened 2022-01-29
Revert "bump-formulae: use `brew bump`."

chenrui333:
This reverts commit 0925dba197462055da7c1fac4cfafa9cb97b9139.

- The `main.rb` is doing more than just livecheck, but also bump the PRs as well (which caused some failed runs like [this](https://github.com/Homebrew/homebrew-core/runs/4989400182))
- We can do another PR of replacing the `livecheck` cmd usage in the `main.rb` script

cc @MikeMcQuaid @dawidd6

### DIFF
--- a/bump-formulae/action.yml
+++ b/bump-formulae/action.yml
@@ -14,8 +14,9 @@ inputs:
 runs:
   using: composite
   steps:
-    - run: brew bump ${{inputs.formulae}}
+    - run: brew ruby $GITHUB_ACTION_PATH/main.rb
       shell: sh
       env:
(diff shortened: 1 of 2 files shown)


dawidd6:
Fixing there:
https://github.com/Homebrew/actions/pull/253
https://github.com/Homebrew/homebrew-core/pull/94084

MikeMcQuaid:
Thanks for the fix @dawidd6!

@chenrui333 Let's try to fix `brew bump` usage ("fix forward") in future rather than reverting ❤️

PR #260, opened 2022-03-02
setup-commit-signing: handle GPG versions < 2.1

Bo98:
This is useful when running in a Homebrew docker container, which won't have gnupg installed by default.

### DIFF
--- a/.github/workflows/setup-commit-signing.yml
+++ b/.github/workflows/setup-commit-signing.yml
@@ -12,7 +12,14 @@ on:
 
 jobs:
   commit-signing:
+    strategy:
+      matrix:
+        container:
+          - ''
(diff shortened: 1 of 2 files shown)


carlocab (approved):
Should we fail early if neither `gpg` nor `brew` is installed/in `PATH`?

Admittedly this is never the case for anything we use it for. But someone might be using this action in a self-hosted runner on GitHub that has neither. Or inside a container that has neither.

Bo98:
Makes sense.

Bo98:
Just realised the docker image _does_ have `gpg`, it's just very old which is why it wasn't working.

I've adjusted this to handle old versions. I've dropped the auto-install since we don't need it for our purpose but will keep the early fail.

PR #262, opened 2022-03-02
setup-commit-signing: allow prespecified GNUPGHOME

Bo98:
The home directory is for some reason not correctly cleared on self-hosted runners (this'll eventually be fixed by ephemeral). For security reasons, allow `GNUPGHOME` to be overridden.


carlocab (approved):

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @Bo98!

PR #263, opened 2022-03-02
setup-commit-signing: set GNUPGHOME in wrapper script

Bo98:
brew env filters `GNUPGHOME`, so make sure it is set within the wrapper script instead of relying on the default.


Bo98:
Oops shouldn't be opening PRs here on a fork.

PR #267, opened 2022-03-29
runtime: upgrade to use nodejs v16

chenrui333:
This is consistent with all the recent github action updates. (mostly in `actions` repos, like `actions/checkout`, `actions/setup-node`, etc.)


Bo98:
There's other changes to do here like upgrading package-lock.json format etc.

Bo98:
Merged in #269.

PR #282, opened 2022-06-05
Replace deprecated `Formula.to_a`

Rylan12:
Remove disabled formulae runs [have not been succeeding](https://github.com/Homebrew/homebrew-core/runs/6741661030?check_suite_focus=true#step:5:23) because `Formula::to_a` is no longer allowed (as of https://github.com/Homebrew/brew/pull/13347). This PR replaces `Formula::to_a` with `Formula::all` to solve the problem.

### DIFF
--- a/remove-disabled-formulae/main.rb
+++ b/remove-disabled-formulae/main.rb
@@ -8,7 +8,7 @@ end
 puts 'Finding disabled formulae...'
 
 one_year_ago = Date.today << 12
-formulae_to_remove = Formula.to_a.select do |formula|
+formulae_to_remove = Formula.all.select do |formula|
   next false unless formula.disabled?
   next false if formula.disable_date.nil?
 


Bo98 (approved):
Thanks!

I wonder if we can do static analysis of deprecations like this at some point to pick up stuff like this.

Rylan12:
What do you mean by "static analysis"?

Bo98:
Like a RuboCop style check so it can flag use of deprecated functions without having to wait until runtime to find out.

Rylan12:
That would be cool but I'm not immediately seeing how to do it. We could probably use RuboCop's parser to figure out which methods have `odeprecated` and `odisabled` calls, but often these are embedded in conditionals since often only a certain use case for the method has been removed rather than the entire thing. I'm not sure how reliable such a check would be

PR #291, opened 2022-08-08
Update bottles failure and output composite action

mohzulfikar:
Based on the PR on Homebrew/homebrew-core#106155 the composite action will fail if we run it on Linux.
This is because the tests on Linux are using custom `workdir`. Thus, we need to somehow pass that directory to the composite action and make the tasks run in the appropriate directory. 
In this PR, I propose the approach to use `inputs`. Any feedback would be appreciated. Thank you.

### DIFF
--- a/failures-summary-and-bottle-result/action.yml
+++ b/failures-summary-and-bottle-result/action.yml
@@ -1,7 +1,7 @@
 name: Failures summary and bottle result
 description: Output the logs of the bottle generation and a summary of all step failures
 inputs:
-  workdir: 
+  workdir:
     description: Working directory of the result
     required: true
     default: ${{ github.workspace }}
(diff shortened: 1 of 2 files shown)


failures-summary-and-bottle-result/action.yml
```diff
@@ -1,17 +1,21 @@
...
       shell: bash
 
     - run: |
-        cat bottles/bottle_output.txt
-        rm bottles/bottle_output.txt
-      shell: bash
-
+        cat ${{ inputs.workdir }}/bottles/bottle_output.txt
+        rm ${{ inputs.workdir }}/bottles/bottle_output.txt
+      shell: bash
```
mistydemeo, line 21: End this file with a newline, please.

mistydemeo (changes requested):

failures-summary-and-bottle-result/action.yml
mohzulfikar: I've added a new line on the new commit. Thank you

mistydemeo (approved):
Thank you!

In the future, it'd be helpful if you make sure to quickly summarize what changes you're making in the pull request title.

PR #315, opened 2022-10-12
Replace deprecated set-output with GITHUB_OUTPUT

bayandin:
Ref https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/remove-disabled-formulae/main.rb
+++ b/remove-disabled-formulae/main.rb
@@ -36,4 +36,7 @@ formulae_to_remove.each do |formula|
   git '-C', tap_dir.to_s, 'commit', formula.path.to_s, '--message', "#{formula.name}: remove formula", '--quiet'
 end
 
-puts '::set-output name=formulae-removed::true'
+github_output = ENV.fetch("GITHUB_OUTPUT") { raise "GITHUB_OUTPUT is not defined" }
+File.open(github_output, "a") do |f|
+  f.puts("formulae-removed=true")
+end
--- a/review-cask-pr/review.rb
+++ b/review-cask-pr/review.rb
@@ -177,8 +177,11 @@ begin
 
       puts message
 
-      puts "::set-output name=event::#{event}"
-      puts "::set-output name=message::#{GitHub::Actions.escape(message)}"
+      github_output = ENV.fetch("GITHUB_OUTPUT") { raise "GITHUB_OUTPUT is not defined" }
+      File.open(github_output, "a") do |f|
+        f.puts("event=#{event}")
+        f.puts("message=#{GitHub::Actions.escape(message)}")
+      end
     end
   when %r{^https://github.com/([^\/]+)/([^\/]+)/pull/(\d+)}
     owner = Regexp.last_match[1]
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -83,8 +83,8 @@ fi
 GEMS_PATH="$HOMEBREW_REPOSITORY/Library/Homebrew/vendor/bundle/ruby/"
 GEMS_HASH="$(shasum -a 256 "$HOMEBREW_REPOSITORY/Library/Homebrew/Gemfile.lock" | cut -f1 -d' ')"
 
-echo "::set-output name=gems-path::$GEMS_PATH"
-echo "::set-output name=gems-hash::$GEMS_HASH"
+echo "gems-path=$GEMS_PATH" >> $GITHUB_OUTPUT
+echo "gems-hash=$GEMS_HASH" >> $GITHUB_OUTPUT
 
 # Setup Homebrew/(home|linux)brew-core tap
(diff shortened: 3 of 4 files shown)


Bo98 (approved):

remove-disabled-formulae/main.rb
```diff
@@ -36,4 +36,7 @@ def git(*args)
   git '-C', tap_dir.to_s, 'commit', formula.path.to_s, '--message', "#{formula.name}: remove formula", '--quiet'
 end
 
-puts '::set-output name=formulae-removed::true'
+github_output = ENV.fetch("GITHUB_OUTPUT") { raise "GITHUB_OUTPUT is not defined" }
```
bayandin, line 39: This will not work, we use `brew ruby` and need to pass $GITHUB_OUTPUT as an argument to the script (the same about other ruby scripts)

https://github.com/Homebrew/actions/blob/bea9c11027209b0a972534bff1fb0b411d51a0ed/remove-disabled-formulae/action.yml#L14

MikeMcQuaid: I don't think we should be relying on `brew ruby` calling `Formula` `GITHUB_OUTPUT`  like this in this repository. Feels like a `brew disabled` or command or `brew info --json --disabled` or something in Homebrew/brew consumed by plain Ruby or Bash would be a better fit.

review-cask-pr/review.rb
```diff
@@ -177,8 +177,11 @@ def review_diff(pr, author_is_member: false)
 
       puts message
 
-      puts "::set-output name=event::#{event}"
-      puts "::set-output name=message::#{GitHub::Actions.escape(message)}"
+      github_output = ENV.fetch("GITHUB_OUTPUT") { raise "GITHUB_OUTPUT is not defined" }
+      File.open(github_output, "a") do |f|
+        f.puts("event=#{event}")
```
MikeMcQuaid, line 182: Same here. This is way more Homebrew API Ruby than we should be having in this repo.

Bo98:
GITHUB_OUTPUT is now whitelisted so this should be ok.

I do not think the usage of Ruby should block this PR at all, given there's no _new_ usages of Ruby here - only pre-existing.

setup-homebrew/main.sh
```diff
@@ -83,8 +83,8 @@ fi
 GEMS_PATH="$HOMEBREW_REPOSITORY/Library/Homebrew/vendor/bundle/ruby/"
 GEMS_HASH="$(shasum -a 256 "$HOMEBREW_REPOSITORY/Library/Homebrew/Gemfile.lock" | cut -f1 -d' ')"
 
-echo "::set-output name=gems-path::$GEMS_PATH"
-echo "::set-output name=gems-hash::$GEMS_HASH"
+echo "gems-path=$GEMS_PATH" >> $GITHUB_OUTPUT
```
dawidd6, line 86: Does this work no issue in our container with non-root user?

MikeMcQuaid (approved):
Fine with me to land this if it works. I'd still rather more of this Ruby logic gets moved but shouldn't block this PR.

setup-homebrew/main.sh
Bo98: Yeah this probably needs a chown

Bo98: I wonder if we can do something here that can work globally for subsequent actions.

Maybe here or something: https://github.com/Homebrew/actions/blob/4019aa33c7aaa82f22b10bdf76fe6ef47b406c61/setup-homebrew/main.sh#L51

MikeMcQuaid: > I wonder if we can do something here that can work globally for subsequent actions.

Good idea 👍🏻

review-cask-pr/review.rb
```diff
@@ -177,8 +177,11 @@ def review_diff(pr, author_is_member: false)
...
+        f.puts("message=#{GitHub::Actions.escape(message)}")
```
Bo98, line 183: Need to check if escaping work the same way (I can check this - just noting here for now).

Bo98:
Will open a separate PR to handle permissions for file commands.

The permission stuff is honestly a bit of a mess. Short of GitHub/Docker properly supporting uid remapping, I had hoped to we could maybe handle this in our Dockerfile by defining an entrypoint that would just change the linuxbrew uid to match what GitHub expects, but it turns out entrypoints are disabled in GitHub Actions unless you have self-hosted with container hooks enabled (why??).

So I guess the cat and mouse game continues...

review-cask-pr/review.rb
Bo98: Answer to this is yes: this needs changing. Not sure if it's best to do here or on the brew side. This might be the only place that uses it for GITHUB_OUTPUT?

PR #317, opened 2022-10-26
setup-homebrew: git config after fixing permissions

jck112:
Without this change the action exits with an error (e.g. `error: could not lock config file`) when running with a token using the _homebrew/ubuntu22.04:master_ image.


MikeMcQuaid (approved):

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @jck112!

PR #320, opened 2022-11-29
create-gcloud-instance: use Ubuntu 22.04

Bo98:
Since we use 22.04 everywhere else. Plus we need it for newer glib.


iMichka (approved):

chenrui333 (commented):
👍

PR #324, opened 2022-12-03
Various improvements to setup-homebrew

Bo98:
Docker file ownership:
* Replace `chown` logic with `setfacl`.
  - This allows the original permissions to not be disrupted, which the runner may need (and does for the new file commands).
* Increase to cover `RUNNER_TEMP`, including file commands (e.g. `GITHUB_OUTPUT`).
* Reset directories to original ACL/permissions after job completes.

Tokens:
* Adjust token behaviour to use HTTP headers rather than URL authentication.
  - This matches how `actions/checkout` behaves.
* Cleanup token config after the job completes.
  - This is a security enchancement.
* Document that tokens persist through job steps.

GitHub workspace:
* Remove `GITHUB_WORKSPACE` replacement for Homebrew/brew and Homebrew/homebrew-core.
  - ⚠️ This is a breaking change affecting those repositories only.
* Adjust `GITHUB_WORKSPACE` handling for other taps to instead clone to the original `GITHUB_WORKSPACE` directory and symlink the tap location to it, rather than vice versa.
  - This has a slight compatibility concern, albeit hopefully not noticeable.
  - This fixes various compatibility issues with actions assuming `GITHUB_WORKSPACE` is untampered.
  - This operation is now cleaned up and undone when the job completes.
  - This behaviour continues (as before) to not apply to container jobs.
* Add `repository-path` output for Homebrew/brew and all taps, pointing to the relevant path matching `GITHUB_REPOSITORY`

Safe directory:
* Whitelist all Homebrew repositories in git's `safe.directory` config when running in a container.
  - This fixes errors when running not as the `linuxbrew` user.
  - This whitelisting action is currently only supported in container jobs.

Other:
* Replace `set-output` with `GITHUB_OUTPUT` file command.
* Move `sbin` to lower priority than `bin`.
  - This makes the behaviour consistent to `brew shellenv`.
* Remove Linux-specific permission changes (experimental - might drop this if `brew tests` fails).

### DIFF
--- a/.github/workflows/setup-homebrew.yml
+++ b/.github/workflows/setup-homebrew.yml
@@ -11,15 +11,42 @@ on:
 jobs:
   setup:
     strategy:
+      fail-fast: false
       matrix:
-        os:
-          - ubuntu-latest
-          - ubuntu-22.04
-          - macOS-latest
-    runs-on: ${{matrix.os}}
+        include:
+          - os: macos-latest
+            key: macos
+          - os: ubuntu-18.04
+            key: ubuntu-18
+          - os: ubuntu-22.04
+            key: ubuntu-22
+          - os: ubuntu-22.04
+            key: docker-root
+            container:
+              image: ghcr.io/homebrew/ubuntu22.04:master
+              options: --user=root
+          - os: ubuntu-22.04
+            key: docker-linuxbrew
+            container:
+              image: ghcr.io/homebrew/ubuntu22.04:master
+              options: --user=linuxbrew
+    runs-on: ${{ matrix.os }}
+    container: ${{ matrix.container }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        run: |
+          # Allow writing under linuxbrew user
+          orig_uid=$(ls -nd $GITHUB_WORKSPACE | awk 'NR==1 {print $3}')
+          sudo chown -R $(whoami) $GITHUB_WORKSPACE $HOME
+
+          git config --global --add safe.directory $GITHUB_WORKSPACE
+          git init
+          git remote add origin $GITHUB_SERVER_URL/$GITHUB_REPOSITORY
+          git fetch --no-tags --depth=1 origin $GITHUB_SHA
+          git checkout --force $GITHUB_SHA
+
+          # Reset ownership so we test if the equivalent in setup-homebrew works.
+          sudo chown -R $orig_uid $GITHUB_WORKSPACE $HOME
 
       - name: Check syntax
         run: bash -n setup-homebrew/*.sh
@@ -33,8 +60,8 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ${{ steps.set-up-homebrew.outputs.gems-path }}
-          key: ${{ runner.os }}-rubygems-${{ steps.set-up-homebrew.outputs.gems-hash }}
-          restore-keys: ${{ runner.os }}-rubygems-
+          key: ${{ matrix.key }}-rubygems-${{ steps.set-up-homebrew.outputs.gems-hash }}
+          restore-keys: ${{ matrix.key }}-rubygems-
 
       - name: Install Homebrew Bundler RubyGems
         if: steps.cache.outputs.cache-hit != 'true'
--- a/setup-homebrew/action.yml
+++ b/setup-homebrew/action.yml
@@ -14,20 +14,17 @@ inputs:
     required: false
     default: false
   token:
-    description: Token to be used for GitHub authentication (if using private repositories)
+    description: Token to be used for GitHub authentication (if using private repositories). This token will persist through other steps for the duration of the job.
     required: false
     default: ''
 outputs:
+  repository-path:
+    description: Path to where the repository has been checked out (if applicable)
   gems-path:
     description: Homebrew's Ruby gems path
-    value: ${{steps.setup.outputs.gems-path}}
   gems-hash:
     description: Homebrew's Ruby Gemfile.lock sha256sum
-    value: ${{steps.setup.outputs.gems-hash}}
 runs:
-  using: composite
-  steps:
-    - run: |
-        "$GITHUB_ACTION_PATH/main.sh" '${{ inputs.test-bot }}' '${{ inputs.debug }}' '${{ inputs.token }}'
-      shell: bash
-      id: setup
+  using: node16
+  main: main.mjs
+  post: post.mjs
--- a/setup-homebrew/main.mjs
+++ b/setup-homebrew/main.mjs
@@ -1,0 +1,14 @@
+import {exec} from "@actions/exec"
+import core from "@actions/core"
+
+// GitHub Actions does not support shell `post` actions and thus requires a JS wrapper.
+try {
+  await exec("bash", [
+    new URL("./main.sh", import.meta.url).pathname,
+    core.getInput("test-bot"),
+    core.getInput("debug"),
+    core.getInput("token")
+  ])
+} catch (error) {
+  core.setFailed(error.message)
+}
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -44,58 +44,93 @@ HOMEBREW_PREFIX="$(brew --prefix)"
 HOMEBREW_REPOSITORY="$(brew --repo)"
 HOMEBREW_CORE_REPOSITORY="$HOMEBREW_REPOSITORY/Library/Taps/homebrew/homebrew-core"
 HOMEBREW_CASK_REPOSITORY="$HOMEBREW_REPOSITORY/Library/Taps/homebrew/homebrew-cask"
+HOMEBREW_OTHER_CASK_REPOSITORIES=(
+    "${HOMEBREW_REPOSITORY}/Library/Taps/homebrew/homebrew-cask-drivers"
+    "${HOMEBREW_REPOSITORY}/Library/Taps/homebrew/homebrew-cask-fonts"
+    "${HOMEBREW_REPOSITORY}/Library/Taps/homebrew/homebrew-cask-versions"
+)
+if [[ "$GITHUB_REPOSITORY" =~ ^.+/homebrew-.+$ ]]; then
+    HOMEBREW_TAP_REPOSITORY="$(brew --repo "$GITHUB_REPOSITORY")"
+fi
 
 # Do in container or on the runner
 if [[ -f "/.dockerenv" ]] || ([[ -f /proc/1/cgroup ]] && grep -qE "actions_job|docker" /proc/1/cgroup); then
-    # Fix permissions to give normal user access
-    sudo chown -R "$(whoami)" "$HOME" "$PWD/.."
+    # Fix permissions to give container user write access.
+    # Directories mounted by the runner are owned by the host user, and can't be written by the container user.
+    if [[ -n "$(command -v setfacl)" ]]; then
+        setfacl_dirs=()
+        user=$(whoami)
+        for dir in "$HOME" "$RUNNER_WORKSPACE" "$RUNNER_TEMP" # These are mounted by the runner
+        do
+            if [[ ! -w "$dir" ]]; then
+                # Give the container user RW permissions, plus execute for directories.
+                sudo setfacl -Rm "d:u:$user:rwX,u:$user:rwX" "$dir"
+                setfacl_dirs+=("$dir")
+            fi
+        done
+        # Store what we've changed so we can revert what we did later.
+        echo "SETFACL_DIRECTORIES=$(IFS=:; echo "${setfacl_dirs[*]}")" >> $GITHUB_STATE
+        echo "Set up ACL."
+    elif [[ ! -w "$GITHUB_OUTPUT" ]]; then
+        # setfacl isn't installed on Ubuntu by default.
+        echo '::warning::Missing write permissions to GitHub directories. Install `acl` (`setfacl`).'
+    fi
+
+    # Add safe directories. This is necessary to allow containers to run as root.
+    # See similar code in `actions/checkout`.
+    # Could do this for non-containers too, but we want to take care to not write into a shared $HOME.
+    # For self-hosted without containers, consider pre-setting this instead.
+    for repo in "$HOMEBREW_REPOSITORY" "$HOMEBREW_CORE_REPOSITORY" \
+                "$HOMEBREW_CASK_REPOSITORY" "$HOMEBREW_OTHER_CASK_REPOSITORIES[@]" \
+                "${HOMEBREW_TAP_REPOSITORY-}"
+    do
+        if [[ -n "$repo" ]]; then
+            git config --global --add safe.directory "$repo"
+        fi
+    done
+
     HOMEBREW_IN_CONTAINER=1
 else
     # Add brew to PATH
-    echo "$HOMEBREW_PREFIX/bin" >> $GITHUB_PATH
     echo "$HOMEBREW_PREFIX/sbin" >> $GITHUB_PATH
+    echo "$HOMEBREW_PREFIX/bin" >> $GITHUB_PATH
 fi
 
 # Use an access token to checkout (private repositories)
 if [[ -n "${TOKEN}" ]]; then
-    git config --global url."https://api:${TOKEN}@github.com/".insteadOf "https://github.com/"
+    git config --global "http.${GITHUB_SERVER_URL}/.extraheader" "Authorization: basic $(echo ${TOKEN} | base64)"
+    echo "TOKEN_SET=1" >> $GITHUB_STATE
 fi
 
 # Setup Homebrew/brew
 if [[ "$GITHUB_REPOSITORY" =~ ^.+/brew$ ]]; then
     cd "$HOMEBREW_REPOSITORY"
-    rm -rf "$GITHUB_WORKSPACE"
-    if [[ -n "${GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED-}" ]]; then
-        mkdir -vp "$GITHUB_WORKSPACE"
-    else
-        ln -vs "$HOMEBREW_REPOSITORY" "$GITHUB_WORKSPACE"
-    fi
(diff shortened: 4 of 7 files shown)


Bo98:
This ideally needs more coverage to actually test the token changes and various update paths involving different taps.

Bo98:
> Remove `GITHUB_WORKSPACE` replacement for Homebrew/brew and Homebrew/homebrew-core.
> * ⚠️ This is a breaking change affecting those repositories only.

This will break most workflows in Homebrew/brew that assume pwd == repo. But it is equally necessary if we want to update from `actions/cache@v1`.

It's maybe worth adding a `repository-path` output here to make it easy to set `working-directory` in subsequent steps.

.github/workflows/setup-homebrew.yml
```diff
@@ -11,15 +11,42 @@ on:
...
+          sudo chown -R $(whoami) $GITHUB_WORKSPACE $HOME
+
+          git config --global --add safe.directory $GITHUB_WORKSPACE
+          git init
+          git remote add origin $GITHUB_SERVER_URL/$GITHUB_REPOSITORY
+          git fetch --no-tags --depth=1 origin $GITHUB_SHA
+          git checkout --force $GITHUB_SHA
+
+          # Reset ownership so we test if the equivalent in setup-homebrew works.
+          sudo chown -R $orig_uid $GITHUB_WORKSPACE $HOME
```
MikeMcQuaid, line 49: Can you explain a bit more what's going on here? I'm pretty lost.

setup-homebrew/main.mjs
```diff
@@ -0,0 +1,13 @@
...
+try {
+  await exec("bash", [
+    new URL("./main.sh", import.meta.url).pathname,
+    core.getInput("test-bot"),
+    core.getInput("debug"),
+    core.getInput("token")
+  ])
+} catch (error) {
+  core.setFailed(error.message)
+}
```
MikeMcQuaid, line 13: I don't really understand why this is needed, can you elaborate a bit?

setup-homebrew/main.sh
```diff
@@ -44,68 +44,96 @@ HOMEBREW_PREFIX="$(brew --prefix)"
...
+            if [[ ! -w "$dir" ]]; then
+                sudo setfacl -Rm "d:u:$user:rwX,u:$user:rwX" "$dir"
+                setfacl_dirs+=("$dir")
+            fi
+        done
+        echo "SETFACL_DIRECTORIES=$(IFS=:; echo "${setfacl_dirs[*]}")" >> $GITHUB_STATE
+        echo "Set up ACL."
+    elif [[ ! -w "$GITHUB_OUTPUT" ]]; then
+        echo "::warning::Missing write permissions to GitHub directories. Install \`acl\` (\`setfacl\`)."
+    fi
```
MikeMcQuaid, line 73: This is pretty involved. Can it get more comments explaining what's going on, why it's needed, etc?

setup-homebrew/main.sh
```diff
@@ -44,68 +44,96 @@ HOMEBREW_PREFIX="$(brew --prefix)"
...
+
+    # Add safe directories.
```
MikeMcQuaid, line 75: What are safe directories? Again: can this get more comments explaining.

setup-homebrew/main.sh
```diff
@@ -138,14 +163,15 @@ else
...
         fi
-        if [[ -z "${HOMEBREW_IN_CONTAINER-}" ]]; then
-            rm -rf "$GITHUB_WORKSPACE"
-            if [[ -n "${GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED-}" ]]; then
-                mkdir -vp "$GITHUB_WORKSPACE"
-            else
-                ln -vs "$HOMEBREW_TAP_REPOSITORY" "$GITHUB_WORKSPACE"
-            fi
+
+        # This is mostly kept for compatibility reasons, so the conditions are intentionally restrictive.
```
MikeMcQuaid, line 167: Can this comment explain more what's going on here

setup-homebrew/post.mjs
```diff
@@ -0,0 +1,11 @@
...
+import core from "@actions/core"
+
+try {
+  await exec("bash", [
+    new URL("./post.sh", import.meta.url).pathname,
+    core.getInput("debug")
+  ])
+} catch (error) {
+  core.setFailed(error.message)
+}
```
MikeMcQuaid, line 11: What's this needed for?

MikeMcQuaid (commented):
Fine with this once it has a few more comments.

Would be great to explain the motivation for the changes rather than just what's changing. I don't really understand the changes here.

Is `$PWD` still the "correct" repository here but it may be a symlink instead? If not: what is `$PWD` now?

Would it be worth splitting this into multiple PRs? This feels like a lot of things to change at once given how many other repositories rely on this.

Bo98:
> Is `$PWD` still the "correct" repository here but it may be a symlink instead? If not: what is `$PWD` now?

`PWD` by default it always `GITHUB_WORKSPACE`. `GITHUB_WORKSPACE` is a fixed location which cannot be changed (but the default working directory can be in the workflow files).

Homebrew and its taps are at fixed locations outside of `GITHUB_WORKSPACE`. To work around the two things being  immovable, we replaced `GITHUB_WORKSPACE` with a symlink to the Homebrew-installed location. This worked for fixing `PWD` but actively broke some official GitHub actions such as `actions/cache@v2` (and later).

We cannot stay on `actions/cache@v1` forever and fight against what GitHub support, and they do not support replacing `GITHUB_WORKSPACE`. The symlink also breaks runner code on non-ephemeral (hence the self-hosted checks that were here).

The alternative is to reverse this so that we always checkout to `GITHUB_WORKSPACE` but symlink on the Homebrew side. As long as Homebrew itself supports symlinks, this should fix all the issues.

Symlinking Homebrew/brew in that direction is however tricky given the nested GitHub repositories. I decided to just drop support for replacing `GITHUB_WORKSPACE` for that repo, given that breakage only affects us and can be addressed in our workflow files. So for Homebrew/brew, `PWD` would no longer point to the repo.

For Homebrew/core, we could keep it. I dropped it considering it was never enabled for self-hosted and our workflows in Homebrew/core need to work on self-hosted, and there is only one thing I think needs changing. I think consistency  between runners makes sense.

For other taps though, I have got the reversed symlink in place with this PR. I have kept the original restrictions of not applying to containers etc. I don't think the restrictions are necessary anymore, but relaxing them does mean making a promise to never reverse that decision, and I want to avoid those promises for now.

.github/workflows/setup-homebrew.yml
Bo98: We never tested setup-homebrew under Homebrew Docker images in this repo's CI, so I thought it was about time we did.

`actions/checkout@v3` does not work with our Homebrew Docker images. The necessary permission fixes are included in our setup-homebrew, but the point of this CI is to checkout so we can test the setup-homebrew changes.

So this replaces `actions/checkout` with a simple checkout + the temporary permission change necessary for that to happen.

setup-homebrew/main.mjs
Bo98: Unfortunantely, GitHub does not support `post` actions in shell so a JS wrapped is needed.

MikeMcQuaid:
> The alternative is to reverse this so that we always checkout to `GITHUB_WORKSPACE` but symlink on the Homebrew side. As long as Homebrew itself supports symlinks, this should fix all the issues.

👍🏻 makes sense.

> So for Homebrew/brew, `PWD` would no longer point to the repo.

I think this is going to cause a bit of confusion, longer-term, and ideally we'd address it. Perhaps a Homebrew/brew environment variable to avoid the symlink-resolving prefix logic would help here?

> For Homebrew/core, we could keep it.

This would be good 👍🏻 

> I think consistency between runners makes sense.

Agreed 👍🏻

setup-homebrew/post.mjs
Bo98: Unfortunantely, GitHub does not support `post` actions in shell so a JS wrapped is needed.

setup-homebrew/main.sh
Bo98: This is the exact same thinng as before with  `chown` but using `setfacl` instead, which works better as it can apply to new files (needed for file commads to work like `GITHUB_OUTPUT`).

`setfacl` _might_ not be installed so I added a check to be safe.

The only other thing happening here is to track what we've changed so that we can cleanup and revert back when the job completes.

setup-homebrew/post.sh
```diff
@@ -0,0 +1,26 @@
...
+
+if [[ -n "${STATE_TOKEN_SET-}" ]]; then
+  git config --global --unset-all "http.${GITHUB_SERVER_URL}/.extraheader"
+  echo "Removed token."
+fi
+
+if [[ -n "${STATE_SETFACL_DIRECTORIES-}" ]]; then
+  (IFS=:; sudo setfacl -Rb ${STATE_SETFACL_DIRECTORIES})
+  echo "Reset permissions."
+fi
```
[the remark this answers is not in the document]
Bo98: This is a cleanup step to restore everything back to how it was before the job started.

* Tokens were previously never cleaned up. This meant subsequent jobs could read them. I consider this to be a security vulnerability.
* Symlinks should be reverted otherwise if the same runnner is used for two taps then we might have symlinks for two taps pointing to the same place.
* `setfacl` reverting I think is good to do. Probably won't _break_ if not done, but I still think it's good to clean up.

setup-homebrew/main.sh
Bo98: Because of https://github.blog/2022-04-12-git-security-vulnerability-announced/, running as root and performing git actions for a directory owned by `linuxbrew` results in an error without this.

See https://github.com/actions/checkout/commit/dcd71f646680f2efd8db4afa5ad64fdcba30e748 for the equivalent change GitHub did for `actions/checkout`.

Bo98:
> I think this is going to cause a bit of confusion, longer-term, and ideally we'd address it. Perhaps a Homebrew/brew environment variable to avoid the symlink-resolving prefix logic would help here?

What type of confusion?

The assumption that PWD == repo already doesn't apply to most of Homebrew/core because that's largely self-hosted.

Note that the code removed was gated under a `Homebrew/brew` check anyway, though that's because we had a bit of code duplication.

MikeMcQuaid:
> > I think this is going to cause a bit of confusion, longer-term, and ideally we'd address it. Perhaps a Homebrew/brew environment variable to avoid the symlink-resolving prefix logic would help here?
> 
> What type of confusion?

Confusion that Homebrew/brew is different to every other repository in this respect and different to what `actions/checkout` does for you.

> The assumption that PWD == repo already doesn't apply to most of Homebrew/core because that's largely self-hosted.

Why is it different there? Can it not be?

setup-homebrew/main.sh
MikeMcQuaid: Sorry, still don't understand what behaviour this changes.

setup-homebrew/post.mjs
MikeMcQuaid: Why do we need post actions now?

setup-homebrew/post.sh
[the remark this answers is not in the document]
MikeMcQuaid: Ok, I see, so this is for cleanup after all of the job has completed and not just `setup-homebrew`.

Good to me, then. Again: would like a lot more inline comments explaining what's going on to future readers.

setup-homebrew/post.mjs
Bo98: See post.sh review below.

Bo98:
> Why is it different there? Can it not be?

Because the `GITHUB_WORKSPACE` replacement broke things. Beyond `actions/checkout`, it broke the GitHub Action's runner's own code on self-hosted.

At this point making PWD == repo would require changes in various places. We currently write to the PWD in test-bot etc and this would break `brew doctor` as a result.

I'm a little hesistent to also do this on non-ephemeral in case the machine crashes mid-job.

setup-homebrew/main.sh
MikeMcQuaid: Thanks! When is this called as root?

Bo98: For our own Docker images: we mostly avoid it ourselves except https://github.com/Homebrew/actions/blob/master/.github/workflows/setup-commit-signing.yml#L20.

For non-Homebrew Docker images: most of them run as root. The `--user=linuxbrew` we have is unusual for the most part (but makes sense for Homebrew/core etc.)

MikeMcQuaid: Ok, thanks.

> GitHub only actually support root in Docker: https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user, so it makes sense to make this work as root.

Note at least internal to GitHub: the opposite is the case 😄.

Given no-one has complained about this I'm fine with including it but don't feel like it's super high priority?

Bo98: > Note at least internal to GitHub: the opposite is the case 😄.

`actions/checkout@v3.1.0` no longer works as non-root anymore. It can't be used with our Homebrew Docker images anymore, hence the `setup-homebrew.yml` CI change.

Bo98:
Added some comments.

.github/workflows/setup-homebrew.yml
MikeMcQuaid: > `actions/checkout@v3` does not work with our Homebrew Docker images. The necessary permission fixes are included in our setup-homebrew, but the point of this CI is to checkout so we can test the setup-homebrew changes.
> 
> So this replaces `actions/checkout` with a simple checkout + the temporary permission change necessary for that to happen.

Could this get added as an inline comment? Thanks ❤️

setup-homebrew/action.yml
```diff
@@ -14,20 +14,17 @@ inputs:
     required: false
     default: false
   token:
-    description: Token to be used for GitHub authentication (if using private repositories)
+    description: Token to be used for GitHub authentication (if using private repositories). This token will persist through other steps for the duration of the job.
```
MikeMcQuaid, line 17:
```suggestion
    description: Token to be used for GitHub authentication (if using private repositories). This token will only persist through other steps for the duration of the job.
```

Maybe?

setup-homebrew/main.sh
```diff
@@ -44,68 +44,101 @@ HOMEBREW_PREFIX="$(brew --prefix)"
...
+)
+if [[ "$GITHUB_REPOSITORY" =~ ^.+/homebrew-.+$ ]]; then
+    HOMEBREW_TAP_REPOSITORY="$(brew --repo "$GITHUB_REPOSITORY")"
+fi
 
 # Do in container or on the runner
 if [[ -f "/.dockerenv" ]] || ([[ -f /proc/1/cgroup ]] && grep -qE "actions_job|docker" /proc/1/cgroup); then
-    # Fix permissions to give normal user access
-    sudo chown -R "$(whoami)" "$HOME" "$PWD/.."
+    # Fix permissions to give container user access.
```
MikeMcQuaid, line 58:
```suggestion
    # Fix permissions to give container user write access.
```

If correct?

setup-homebrew/main.sh
```diff
@@ -44,68 +44,101 @@ HOMEBREW_PREFIX="$(brew --prefix)"
...
+                sudo setfacl -Rm "d:u:$user:rwX,u:$user:rwX" "$dir"
+                setfacl_dirs+=("$dir")
+            fi
+        done
+        # Store what we've changed so we can revert what we did later.
+        echo "SETFACL_DIRECTORIES=$(IFS=:; echo "${setfacl_dirs[*]}")" >> $GITHUB_STATE
+        echo "Set up ACL."
+    elif [[ ! -w "$GITHUB_OUTPUT" ]]; then
+        # setfacl isn't installed on Ubuntu by default.
+        echo "::warning::Missing write permissions to GitHub directories. Install \`acl\` (\`setfacl\`)."
```
MikeMcQuaid, line 76:
```suggestion
        echo '::warning::Missing write permissions to GitHub directories. Install `acl` (`setfacl`).'
```

setup-homebrew/main.sh
```diff
@@ -138,14 +168,16 @@ else
...
-        if [[ -z "${HOMEBREW_IN_CONTAINER-}" ]]; then
-            rm -rf "$GITHUB_WORKSPACE"
-            if [[ -n "${GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED-}" ]]; then
-                mkdir -vp "$GITHUB_WORKSPACE"
-            else
-                ln -vs "$HOMEBREW_TAP_REPOSITORY" "$GITHUB_WORKSPACE"
-            fi
+
+        # Make repo available under `GITHUB_WORKSPACE` (default working directory), which some third-party taps may need.
+        # The symlink needs to be in this direction or `actions/cache` etc will break.
```
MikeMcQuaid, line 173:
```suggestion
        # The symlink needs to be in this direction or `actions/cache` etc. will break as they rely on `GITHUB_WORKSPACE` being `PWD`
```

or whatever is correct

setup-homebrew/action.yml
Bo98: Possibly, but the tone of this is intended to be more of a security warning, to let the user know that the token could be read by subsequent steps (in case there's any untrusted code being run), rather than it being scoped to just setup-homebrew.

Bo98: 1) Yes, with an opt-out (`persist-credentials`). Though they also apply the token by default unlike us.
2) action.yml states `The post-job step removes the PAT.`. The better description is at the near the start of the README: https://github.com/actions/checkout/blob/main/README.md

MikeMcQuaid: Fine with me either way. Let's emulate what they do as much as possible, I reckon, because we're sort of a replacement for that for taps.

carlocab (approved):

Bo98:
Will merge this when I'm sure I'll be around to monitor CI afterwards. Today's the last day of the holidays (in Scotland) so probably tomorrow.

chenrui333 (approved):

PR #331, opened 2023-02-10
setup-homebrew: fix token handling

### LINKED ISSUE #330: Changes in #324 break our actions/setup-homebrew deployment
Hi @Bo98 , I've read the changes in #324 , and it seems the way this change:

> Adjust token behaviour to use HTTP headers rather than URL authentication.
> This matches how actions/checkout behaves.

Changed how token authentication works broke the actions ability to clone a repo in the setup-homebrew action.

This is the error we're getting:
```
fatal: could not read Username for 'https://github.com/': No such device or address
```

Our setup has not changed since the last time we ran a test-bot workflow, here's what our invocation looks like:
```
      - name: Set up Homebrew
        id: set-up-homebrew
        uses: Homebrew/actions/setup-homebrew@master
        with:
          token: ${{ secrets.HOMEBREW_GITHUB_API_TOKEN }} 
```

Please let me know if there's any additional environmental variables we have to set up to get this working. Cheers!

Bo98:
And mask it because it's a secret.

Fixes #330.

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -98,7 +98,9 @@ fi
 
 # Use an access token to checkout (private repositories)
 if [[ -n "${TOKEN}" ]]; then
-    git config --global "http.${GITHUB_SERVER_URL}/.extraheader" "Authorization: basic $(echo ${TOKEN} | base64)"
+    base64_token=$(echo -n "x-access-token:${TOKEN}" | base64)
+    echo "::add-mask::${base64_token}"
+    git config --global "http.${GITHUB_SERVER_URL}/.extraheader" "Authorization: basic ${base64_token}"
     echo "TOKEN_SET=1" >> $GITHUB_STATE
 fi
 


setup-homebrew/main.sh
```diff
@@ -98,7 +98,9 @@ fi
 
 # Use an access token to checkout (private repositories)
 if [[ -n "${TOKEN}" ]]; then
-    git config --global "http.${GITHUB_SERVER_URL}/.extraheader" "Authorization: basic $(echo ${TOKEN} | base64)"
+    base64token=$(echo -n "x-access-token:${TOKEN}" | base64)
```
MikeMcQuaid, line 101:
```suggestion
    local base64_token=$(echo -n "x-access-token:${TOKEN}" | base64)
```

maybe?

MikeMcQuaid (approved):

ryantimjohn:
Thanks for the quick turnaround on this @Bo98, will try and run our GHA and report out here.

ryantimjohn:
@Bo98 I see this PR as still open, are there any updates here?

Bo98:
Oh sorry, misread your comment as you were going to test this. Will get this merged in just a moment.

setup-homebrew/main.sh
Bo98: Added the underscore, but for `local`:

> main.sh: line 101: local: can only be used in a function

ryantimjohn:
Apologies, that makes perfect sense. I was going to run you'd merged but could've just run from your commit. Sorry about that.

PR #332, opened 2023-02-17
setup-homebrew: handle installs without Homebrew/core tapped

Bo98:
For CI outside of Homebrew/core, I've set cloning to be gated around `HOMEBREW_NO_INSTALL_FROM_API`. Core cloning is very slow so I'm trying to avoid there where possible.

### DIFF
--- a/.github/workflows/setup-homebrew.yml
+++ b/.github/workflows/setup-homebrew.yml
@@ -71,4 +71,4 @@ jobs:
 
       - run: brew test-bot --only-setup
 
-      - run: brew test-bot --only-tap-syntax
+      - run: brew info hello
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -130,8 +130,15 @@ echo "gems-hash=$GEMS_HASH" >> $GITHUB_OUTPUT
 
 # Setup Homebrew/(home|linux)brew-core tap
 if [[ "$GITHUB_REPOSITORY" =~ ^.+/(home|linux)brew-core$ ]]; then
-    cd "$HOMEBREW_CORE_REPOSITORY"
-    git remote set-url origin "https://github.com/$GITHUB_REPOSITORY"
+    if [[ -d "$HOMEBREW_CORE_REPOSITORY" ]]; then
+        cd "$HOMEBREW_CORE_REPOSITORY"
+        git remote set-url origin "https://github.com/$GITHUB_REPOSITORY"
+    else
+        mkdir -vp "$HOMEBREW_CORE_REPOSITORY"
+        cd "$HOMEBREW_CORE_REPOSITORY"
+        git init
+        git remote add origin "https://github.com/$GITHUB_REPOSITORY"
+    fi
     git_retry fetch origin "$GITHUB_SHA" '+refs/heads/*:refs/remotes/origin/*'
     git remote set-head origin --auto
     git checkout --force -B master FETCH_HEAD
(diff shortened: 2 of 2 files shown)


Bo98:
Merging to unblock CI (hopefully).

scpeters:
thanks @Bo98; this fixes the `Set up Homebrew` step for me, but I'm still getting failures when checking the syntax of my 3rd-party tap as it can't find dependencies from `homebrew-core`:

* https://github.com/osrf/homebrew-simulation/pull/2216: https://github.com/osrf/homebrew-simulation/actions/runs/4206857286/jobs/7304280359

~~~
[14](https://github.com/osrf/homebrew-simulation/actions/runs/4206857286/jobs/7304280359#step:8:15)
==> brew audit --tap=osrf/simulation
==> FAILED
Full audit --tap=osrf/simulation output
  osrf/simulation/bullet@2.87:
    * Can't find dependency '"cmake"'.
...
~~~

scpeters:
I'll try tapping homebrew-core explicitly...

scpeters:
> I'll try tapping homebrew-core explicitly...

ok that worked for me in https://github.com/osrf/homebrew-simulation/pull/2217

Bo98:
Hmm ok, that's useful info.

I've avoided tapping Homebrew/core by default because it is a slow thing to enforce and not everyone needs it.

We could enforce it in test-bot, though I feel the audit is the buggy part here. Non-core taps should be able to work with an API core I reckon.

Bo98:
Ok I reckon it's because we auto set `HOMEBREW_NO_INSTALL_FROM_API` for `brew audit` but that's wrong if we're auditing a non-core formula.

0xbabo:
I ran into this yesterday after creating a new tap and scratched my head over it for a while. Fixed it today by making the formulae depend explicitly on `homebrew/core/<formula>` (https://github.com/0xbabo/homebrew-nostr/commit/cda9dea004525030c3c6ff8de06636eb42d23c48) rather than having the CI tap the whole `homebrew/core`. Is that a reasonable workaround or might it cause some other issues?

Bo98:
Interesting spot. Looks like the audit does not print errors if it can't find a formula specified with a tap.

Workaround sounds reasonable. Shouldn't cause any other obvious issues as far as I know.

.github/workflows/setup-homebrew.yml
```diff
@@ -71,4 +71,4 @@ jobs:
 
       - run: brew test-bot --only-setup
 
-      - run: brew test-bot --only-tap-syntax
+      - run: brew info hello
```
MikeMcQuaid, line 74: Why did this change? Want to make sure we're still testing the `brew test-bot` case here?

setup-homebrew/main.sh
```diff
@@ -130,8 +130,15 @@ echo "gems-hash=$GEMS_HASH" >> $GITHUB_OUTPUT
 
 # Setup Homebrew/(home|linux)brew-core tap
 if [[ "$GITHUB_REPOSITORY" =~ ^.+/(home|linux)brew-core$ ]]; then
```
MikeMcQuaid, line 132: Can probably remove all the `linuxbrew-core` logic now.

setup-homebrew/main.sh
```diff
@@ -187,16 +194,20 @@ else
...
     fi
 
-    # Migrate linuxbrew-core to homebrew-core
-    HOMEBREW_CORE_REPOSITORY_ORIGIN="$(git -C "$HOMEBREW_CORE_REPOSITORY" remote get-url origin 2>/dev/null)"
-    if [[ "$HOMEBREW_CORE_REPOSITORY_ORIGIN" == "https://github.com/Homebrew/linuxbrew-core" ]]
-    then
-        git -C "$HOMEBREW_CORE_REPOSITORY" remote set-url origin "https://github.com/Homebrew/homebrew-core"
-    fi
+    if [[ -d "${HOMEBREW_CORE_REPOSITORY}" ]]; then
+        # Migrate linuxbrew-core to homebrew-core
```
MikeMcQuaid, line 198: This can probably go away now?

MikeMcQuaid:
> Given this one audit is the only thing affected, I say it should be fixed on the audit side.

Agreed 👍🏻

.github/workflows/setup-homebrew.yml
Bo98: Tap syntax checks require a Homebrew/core clone (and are not the quickest).

We already test cleanup and setup parts of test-bot here, which should cover if `brew` is set up correctly. But we could add `--test-default-formula` or something.

MikeMcQuaid: > But we could add --test-default-formula or something if you feel more is needed.

Yeh, that'd be nice I think 👍🏻 

> Tap syntax checks require a Homebrew/core clone, which doesn't seem worth adding to what is already a long step taking several minutes.

Agreed 👍🏻

PatTheMav:
The workflows generated by `brew tap-new` still contain the `brew test-bot --only-tap-syntax` step, which will obviously break for the same reason. Which repo would be the most appropriate to report that in?

Bo98:
`brew test-bot --only-tap-syntax` should take your current tap rather than Homebrew/core.

PatTheMav:
> Ok it's because we auto set `HOMEBREW_NO_INSTALL_FROM_API` for `brew audit` but that's wrong if we're auditing a non-core formula. If we don't do that then the whole of test-bot works fine.
> 
> Given this one audit is the only thing affected, I say it should be fixed on the audit side.

Manually tapping Homebrew/core is still required to get the actions on a custom tap working - ist it because the above hasn't been fixed yet?

Bo98:
Depends what error you get exactly. If you are getting a missing dependency error then yeah it'll be that.

PatTheMav:
That is correct - the test workflow breaks because it cannot find the dependencies specified in formulas without tapping core.

I'm currently running into rate limiting issues with the GitHub API, but will report back if the publish workflow also has any additional issues.

Bo98:
I've pushed something that will hopefully fix this.

PatTheMav:
> I've pushed something that will hopefully fix `brew audit`.

Thanks - I just noticed that the "publish" workflow also has the same issue:

```
* Can't find dependency '"ncurses"'.
  * Can't find dependency '"gettext"'.
  * Can't find dependency '"lua"'.
  * Can't find dependency '"luajit"'.
```

Bo98:
Yeah publish probably runs an audit internally somewhere.

PR #336, opened 2023-03-17
Replace deprecated `set-output` command

nandahkrishna:
See here: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

TL;DR: The `set-output` command has been deprecated and we should be writing to the `$GITHUB_OUTPUT` environment file instead.

Old syntax:
```
- name: Set output
  run: echo "::set-output name={name}::{value}"
```

New syntax:
```
- name: Set output
  run: echo "{name}={value}" >> $GITHUB_OUTPUT
```

### DIFF
--- a/review-cask-pr/review.rb
+++ b/review-cask-pr/review.rb
@@ -177,8 +177,10 @@ begin
 
       puts message
 
-      puts "::set-output name=event::#{event}"
-      puts "::set-output name=message::#{GitHub::Actions.escape(message)}"
+      File.open(ENV["GITHUB_OUTPUT"], "a") do |f|
+        f.puts("event=#{event}")
+        f.puts(GitHub::Actions.format_multiline_string("message", message))
+      end
     end
   when %r{^https://github.com/([^\/]+)/([^\/]+)/pull/(\d+)}
     owner = Regexp.last_match[1]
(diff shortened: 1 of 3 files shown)


review-cask-pr/review.rb
```diff
@@ -177,8 +177,10 @@ def review_diff(pr, author_is_member: false)
 
       puts message
 
-      puts "::set-output name=event::#{event}"
-      puts "::set-output name=message::#{GitHub::Actions.escape(message)}"
+      File.open(ENV["GITHUB_OUTPUT"], "a") do |f|
+        f.puts("event=#{event}")
+        f.puts("message=#{GitHub::Actions.escape(message)}")
```
Bo98, line 182: Escaping works differently now so this doesn't work. The new way to do it is:

```
{name}<<{delimiter}
{value}
{delimiter}
```

For security reasons `delimiter` must be randomly generated and unique to each run (e.g. UUID v4). It should also be something that doesn't appear in the `message`. It may be worth having a new method in `Homebrew/brew`.

See https://github.com/actions/toolkit/blob/b00a9fd033f4b30f2355acd212f531ecbbb9b38f/packages/core/src/file-command.ts#L27-L47 for an example implementation in JavaScript.

Bo98 (commented):
Needs one adjustment before it can be merged but this otherwise looks good.

nandahkrishna:
Merging since the comment was addressed.

Bo98:
Looks good, thanks!

PR #343, opened 2023-03-31
Don't count deleted files for `missing_content` constraint.

reitermarkus:
Deleted files always match all `missing_content` constraints, adding superfluous labels.


dawidd6 (approved):

MikeMcQuaid (approved):

MikeMcQuaid:
Makes sense, thanks @reitermarkus!

PR #369, opened 2023-05-09
setup-homebrew: remove reliance on a functional brew Ruby

Bo98:
Just like how `brew update` is able to work if the Ruby side is completely broken, `setup-homebrew` should work similarly to get a functional version.

https://github.com/Homebrew/homebrew-core/actions/runs/4920922528/jobs/8790636214#step:2:37
(in this case, the bad version was from a Homebrew/brew CI run - not something that was merged)

In the longer term, we should perhaps implement `brew --repo <tap>` in shell.


Bo98:
CI failure is caused by https://github.com/Homebrew/homebrew-test-bot/commit/ce58be9afc9e4fb823e88a64a93c536834693378 - global gitconfigs are no longer respected after that change.

MikeMcQuaid:
> In the longer term, we should perhaps implement `brew --repo <tap>` in shell.

I like this idea 👍🏻 (at least for the typical, GitHub, not custom origin case).

PR #377, opened 2023-06-08
remove cask-drivers from setup-homebrew/main.sh

razvanazamfirei:
remove cask-drivers from setup-homebrew/main.sh


MikeMcQuaid (approved):
Thanks again @razvanazamfirei!

PR #381, opened 2023-06-29
setup-homebrew: improve progress logging

Bo98:
The only reason I can read and understand the setup-homebrew logs is because I know the order of operations.

Let's print whenever we start each fetch.


MikeMcQuaid (approved):
Great idea

PR #382, opened 2023-07-02
setup-homebrew: always update cask repos

Bo98:
Instead of only in homebrew-cask* CI.

Seems like the right thing to do if it isn't too much slower.


MikeMcQuaid:
Thanks @Bo98!

PR #383, opened 2023-07-03
dependabot.yml: update all dependency types

Bo98:
The docs claim this does nothing on npm, but I want to see if that's still true.


Bo98:
> The docs claim this does nothing on npm

And all the new PRs suggest otherwise :-)

PR #391, opened 2023-07-04
setup-homebrew: add ability to opt-out of tap updates

Bo98:
Idea is this could make the `formulae.brew.sh` workflow a little more reliable by making the homebrew-core update job opt-out of updating the homebrew-cask tap (`cask: false`) and vice versa for the homebrew-cask update job.

Default behaviour continues to be to update both, since that's the safest and most reasonable thing to do.

### DIFF
--- a/setup-homebrew/action.yml
+++ b/setup-homebrew/action.yml
@@ -5,6 +5,14 @@ branding:
   icon: settings
   color: yellow
 inputs:
+  core:
+    description: Update the `homebrew/core` tap.
+    required: false
+    default: true
+  cask:
+    description: Update the `homebrew/cask` tap.
+    required: false
+    default: true
   test-bot:
     description: Install the `homebrew/test-bot` tap.
     required: false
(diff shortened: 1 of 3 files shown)


bevanjkay:
@Bo98 Is there any reason these shouldn't be set to `false` in a third-party tap for CI runs (aside from potential errors/warnings when formula/casks set to `conflicts_with` are missing)?

Bo98:
That plus we don't actually know what is going to happen for the rest of the workflow (there might be a `brew install --cask`).

Idea is that `setup-homebrew` works like how `brew update` does

MikeMcQuaid (approved):
Makes sense to me, thanks @Bo98!

MikeMcQuaid:
https://github.com/Homebrew/actions/actions/runs/5449628267/jobs/9923314007?pr=391#step:8:6 seems broken unrelated to this PR.

Bo98:
> https://github.com/Homebrew/actions/actions/runs/5449628267/jobs/9923314007?pr=391#step:8:6 seems broken unrelated to this PR.

Yeah issue on the test-bot side I think, particularly since it has gotten that far into the workflow.

PR #394, opened 2023-07-06
setup-homebrew/main.sh: unset some env vars before `brew update`.

MikeMcQuaid:
This should avoid repeatedly updating the same taps e.g. https://github.com/Homebrew/brew/pull/15634#issuecomment-1623885350


Bo98 (approved):

chenrui333 (commented):
👍

PR #397, opened 2023-07-07
setup-homebrew: fixes for brew and homebrew-core

Bo98:
* Allow `HOMEBREW_NO_INSTALL_FROM_API` for homebrew-core.
* Set `HOMEBREW_UPDATE_SKIP_BREW` which I'll use in brew to skip updating brew since we already do it here.


MikeMcQuaid:
Thanks @Bo98!

PR #398, opened 2023-07-08
setup-homebrew: skip autoupdate for formulae.brew.sh CI

Bo98:
Otherwise you have a chicken and egg problem where generating formulae.brew.sh requires a working formulae.brew.sh.

Usually is fine but in the event formulae.brew.sh has issues then it shouldn't prevent us from deploying.

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -144,15 +144,17 @@ else
     fi
 fi
 
-# Run `brew update` once to e.g. download formula/cask JSON files.
-(
-    ohai "brew update --auto"
-    # Unset these to ensure homebrew/core and homebrew/cask are not updated (again)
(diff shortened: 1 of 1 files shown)


Bo98:
> * are there any other repos where we want to do this?

It's maybe not strictly necessary for homebrew-core (unless we have a tap_migrations audit?), but also doesn't break anything either.

It was desirable for formulae.brew.sh CI as that CI should not fail if formulae.brew.sh is not working as it is responsible for pushing new changes to formulae.brew.sh.

> should/could this be a Actions configuration variable instead that defaults to `true`?

If we want it in any other repo other than formulae.brew.sh, then yes possibly. Though I reckon only if such a config allowed `HOMEBREW_NO_INSTALL_FROM_API` too otherwise we're going to be getting into situations where we having a version of brew several weeks newer than the JSON file.

MikeMcQuaid:
@Bo98 Cool, makes sense to me, I'll add a comment and that's enough 👍🏻

PR #411, opened 2023-09-04
feat: bump runtime to node 20

chenrui333:
Node12 was deleted from runner. Node20 was added to Actions Runner on v2.308.0.
Node16 has en end of life on 11 Sep 2023.

This PR updates the default runtime to node20, rather then node16

---

also bump checkout action to v4

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -32,7 +32,7 @@ jobs:
           cask: false
           test-bot: false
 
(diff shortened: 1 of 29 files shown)


dawidd6:
You would need to switch from fork PR to origin repo PR for all the checks to work properly.

MikeMcQuaid:
@chenrui333 please reopen from a non-fork, thanks!

PR #416, opened 2023-09-19
Some updates

chenrui333:
- docs: bump cache action to v3
- review-cask-pr: add .ruby-version and update workflow


Bo98:
`review-cask-pr` uses `brew ruby` so it shouldn't require any external Ruby.

.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 3.2.2
```
MikeMcQuaid, line 1: What tool consumes these files?

MikeMcQuaid (commented):
Looks good other than one question.

.tool-versions
chenrui333: the setup-ruby would consume this (i had some issue putting it into review-cask-pr folder, which is kind of a bug in setup-ruby action)

chenrui333: gonna merge for now, we can discuss/revisit if needed

PR #418, opened 2023-09-26
check-commit-format: fail the status check when autosquash is needed

ZhongRuoyu:
Contributors do not generally seem to understand the implication of the `autosquash` label, and we often need to ask them to squash the commits, do the work for them, or let the PR be replaced (which is less ideal as it requires another full round of CI run). Let's instead fail this check and make the status check message more explicit, hoping this will draw more attention.


MikeMcQuaid (approved):

MikeMcQuaid:
Makes sense, thanks @ZhongRuoyu! I think the ideal outcome would be that the autosquash (force) pushes to the original PR so that the workflow is the same for everyone whether BrewTestBot pushes the squashed commit, a contributor or maintainer does.

PR #426, opened 2023-10-13
setup-homebrew: retry `git remote set-head` calls

Bo98:
Mitigation for GitHub networking issues

`git remote set-head` documentation states:

> With `-a` or `--auto`, **the remote is queried** to determine its `HEAD`, then [...]

This is a network request that could fail if there are networking issues so wrap it in a retry too.


MikeMcQuaid:
Also makes a lot of sense to have indefinitely, thanks @Bo98!

PR #427, opened 2023-10-14
setup-homebrew: don't reset test-bot to master in test-bot repo

Bo98:
Fixes https://github.com/Homebrew/homebrew-test-bot/pull/975#issuecomment-1762500482


MikeMcQuaid:
Good catch, thanks @Bo98!

PR #458, opened 2023-11-17
Disable base64 line wrapping on Linux.

### LINKED ISSUE #457: GitHub authorization header construction can fail on Linux
If the setup-homebrew action is configured with a custom GitHub token, then when the action encodes it into a basic authorization header using [base64](https://man7.org/linux/man-pages/man1/base64.1.html), it can end up being split into multiple lines because the default Linux behavior is to wrap the output at 76 columns. This results in a broken Authorization header, and GitHub access failing with HTTP 400 responses.

jpeach:
The GNU coreutils base64 command will wrap long tokens to 74-character lines, breaking the Authorization header. Disable line wrapping on Linux.

This fixes #457.

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -123,7 +123,8 @@ fi
 
 # Use an access token to checkout (private repositories)
 if [[ -n "${TOKEN}" ]]; then
-    base64_token=$(echo -n "x-access-token:${TOKEN}" | base64)
+    # GNU coreutils will line-wrap long tokens, so delete any newlines.
+    base64_token=$(echo -n "x-access-token:${TOKEN}" | base64 | tr -d "\n")
     echo "::add-mask::${base64_token}"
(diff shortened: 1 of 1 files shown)


jpeach:
cc @Bo98

Bo98:
Ah, that's annoying. Thanks for the fix!

Since it's possible to install GNU base64 on macOS, I wonder if the most cross-platform fix here is to simply `| tr -d "\n"`

jpeach:
> Ah, that's annoying. Thanks for the fix!
> 
> Since it's possible to install GNU base64 on macOS, I wonder if the most cross-platform fix here is to simply `| tr -d "\n"`

Yeh that works on both Ubuntu an macOS. Lemme update the PR.

MikeMcQuaid (approved):
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @jpeach!

For future reference: no need to open an issue if you're going to immediately create a PR to address it. I know other projects prefer this but: we don't. All good, you're not to know!

PR #478, opened 2023-12-30
label-pull-requests: add `allow_any_match`

cho-m:
Not sure about naming.

Currently need a way to correctly label PRs when they modify more than one file (e.g. alias, multiple synced formulae, etc).

Particularly for some CI-impacting labels which can be problematic if missing, e.g. `long build` & `CI-build-dependents-from-source` for `ghc` always missing on next major/minor release.

### DIFF
--- a/label-pull-requests/README.md
+++ b/label-pull-requests/README.md
@@ -18,7 +18,8 @@ Will remove labels from a pull request that no longer apply.
           {
               "label": "new formula",
               "status": "added",
-              "path": "Formula/.+"
+              "path": "Formula/.+",
+              "allow_any_match": true
           },
           {
               "label": "bottle unneeded",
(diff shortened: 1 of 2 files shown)


MikeMcQuaid (approved):
Makes sense to me, thanks @cho-m!

cho-m:
Still don't know how CI checks work for actions repo.

Bo98:
It basically doesn't work for forked repos at the moment. I'm still working on how to best handle that - there's not really one global answer here.

You should however have privileges to push to a branch here however and open a PR

chenrui333:
yeah, CI run failure is because of the repo fork issue.

MikeMcQuaid:
Merging and we can revert if there's issues.

PR #483, opened 2024-01-26
Add node_modules to .gitignore file

ahmdyasser:
node_modules folder is generated when install JS packages, it shouldn't be part of the repo.


Bo98:
It is necessary for the GitHub Actions to run, as shown by the failing CI here. The alternative is bundling the dependencies into a `dist` file (like how GitHub's own actions work) but is ultimately the same result.

ahmdyasser:
@Bo98 Got it now, thank you

PR #488, opened 2024-02-15
actions: use `/bin/bash`

ZhongRuoyu:
When there is a broken `bash` in `PATH`, some actions may fail as in Homebrew/homebrew-core#162819. (See [this job](https://github.com/Homebrew/homebrew-core/actions/runs/7916916851/job/21612022470?pr=162819).) Let's modify those actions to use `/bin/bash` instead of an arbitrary `bash` from `PATH`; `brew` requires and always runs on `/bin/bash` anyway.

### DIFF
--- a/count-bottles/action.yml
+++ b/count-bottles/action.yml
@@ -27,5 +27,5 @@ runs:
   steps:
     - run: ./count.sh '${{ inputs.working-directory }}' '${{ inputs.debug }}'
       working-directory: ${{ github.action_path }}
-      shell: bash
+      shell: /bin/bash -euo pipefail {0}
       id: count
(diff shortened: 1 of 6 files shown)


Bo98:
As a slight nit: `-euo pipefail` is more like what we had before but LGTM.

count-bottles/action.yml
```diff
@@ -27,5 +27,5 @@ runs:
   steps:
     - run: ./count.sh '${{ inputs.working-directory }}' '${{ inputs.debug }}'
       working-directory: ${{ github.action_path }}
-      shell: bash
+      shell: /bin/bash -e {0}
```
MikeMcQuaid, line 30:
```suggestion
      shell: /bin/bash -euo pipefail {0}
```

may want to consider that on all of these, assuming it doesn't break them.

MikeMcQuaid (approved):
Great move, one suggestion (that can be a later PR).

ZhongRuoyu:
Thanks for the quick review @Bo98 @MikeMcQuaid!

PR #497, opened 2024-02-25
Revert "bump-packages: use `--no-fork` switch on our taps"

p-linnane:
This needs some changes on the `brew` side to handle this, so reverting until that can be addressed.


p-linnane:
Confirming we are back to a working state: https://github.com/Homebrew/homebrew-cask/actions/runs/8040564108/job/21958806049

PR #499, opened 2024-02-28
limit-pull-requests: do not use `gh pr list`

ZhongRuoyu:
`gh pr list` does not list any PRs from users with a private profile (cli/cli#6879), so let's use another method to count PRs.

This allows us to have a limit higher than 99 too, though not quite necessary.

Before:

```console
$ gh pr list --state=open --author=BrewTestBot --json=number --jq=length --limit=100
23
```

After:

```console
$ gh api \
  --method GET \
  --header 'Accept: application/vnd.github+json' \
  --header 'X-GitHub-Api-Version: 2022-11-28' \
  --field state=open \
  --paginate \
  '/repos/Homebrew/homebrew-core/pulls' |
  jq \
    --raw-output \
    --arg USER 'BrewTestBot' \
    'map(select(.user.login == $USER)) | length'
23
```

### DIFF
--- a/limit-pull-requests/action.yml
+++ b/limit-pull-requests/action.yml
@@ -21,7 +21,7 @@ inputs:
   comment-limit:
     description: >
       Post the comment when the user's number of open pull requests exceeds this
-      number and `comment` is not empty (maximum: 99)
+      number and `comment` is not empty
     required: true
     default: "10"
   comment:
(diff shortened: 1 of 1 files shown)


Bo98:
Honestly we probably should just use Octokit.js instead in actions like this as it has much less quirks like this.

Bo98 (approved):

PR #501, opened 2024-02-28
limit-pull-requests: convert strings to numbers before comparing them

ZhongRuoyu:
Otherwise we'd see funny results.

See https://docs.github.com/en/actions/learn-github-actions/expressions#operators.


MikeMcQuaid:
Good catch, thanks @ZhongRuoyu!

PR #502, opened 2024-02-29
limit-pull-requests: better handle quotes in comment body

ZhongRuoyu:
By putting the comment body in an environment variable we can avoid having to escape quotes at all.


Bo98:
Thanks!

chenrui333 (commented):
Nice job!

MikeMcQuaid:
Good idea, thanks @ZhongRuoyu!

PR #506, opened 2024-03-01
Set `HOMEBREW_TEST_BOT_AUTOBUMP` in bump-packages

apainintheneck:
This will unblock https://github.com/Homebrew/brew/pull/16750 by providing us with an environment variable that's only on when we're autobumping packages. In all other scenarios, we want people to get an error message if they try to bump one of the packages that gets regularly autobumped on CI.

### DIFF
--- a/bump-packages/action.yml
+++ b/bump-packages/action.yml
@@ -29,6 +29,7 @@ runs:
       if: inputs.formulae != ''
       env:
         HOMEBREW_DEVELOPER: "1"
+        HOMEBREW_TEST_BOT_AUTOBUMP: "1"
         HOMEBREW_GITHUB_API_TOKEN: ${{ inputs.token }}
     - run: |
         if [[ "${GITHUB_REPOSITORY_OWNER}" == 'Homebrew' ]]; then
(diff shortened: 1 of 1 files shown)


Bo98:
Looks good but this can only be merged after the current incorrect usage of it in brew (dupe PR checking) is removed.

apainintheneck:
> Looks good but this can only be merged after the current incorrect usage of it in brew (dupe PR checking) is removed.

I just realized that.

I guess we need three PRs to be merged in this order, right?

1. To drop the incorrect usage in Core.
2. This one.
3. https://github.com/Homebrew/brew/pull/16750

Bo98:
Yes

ZhongRuoyu:
Do we want to set this in the autobump action in core and cask instead? There is some usage of this action outside of Homebrew, where it doesn't seem necessary/reasonable to set `HOMEBREW_TEST_BOT_AUTOBUMP`.

https://github.com/search?q=Homebrew%2Factions%2Fbump-packages&type=code
https://github.com/search?q=Homebrew%2Factions%2Fbump-formulae&type=code

apainintheneck:
I think those are mostly forks but it's probably to scope this down as much as possible anyway. I'll close this PR and then we can just add it to each repo.

PR #511, opened 2024-03-18
label-pull-requests: `allow_any_match` only when adding label

cho-m:
This might be reason for bug where `constraint.wanted == false` is removing the label even though there is a file that matches and `allow_any_match == true`


MikeMcQuaid:
Thanks @cho-m!

PR #513, opened 2024-03-22
setup-homebrew: support stable brew

cho-m:
Want to add a job to `homebrew/core` to check tap_syntax on stable tag.

Currently, we need to manually keep track of this to avoid accidentally introducing errors to users on stable tags.


MikeMcQuaid (approved):
Makes sense, thanks again @cho-m!

I wonder if this should be default outside the Homebrew org 🤔

PR #538, opened 2024-05-17
setup-homebrew: handle stable on stable Docker images

Bo98:
* Don't assume `origin/HEAD` exists
* Don't checkout master first for stable - we don't support downgrades.


MikeMcQuaid (approved):
Makes sense, thanks @Bo98!

PR #541, opened 2024-05-22
workflows/git-try-push: use local directory

Bo98:
Avoids the need for a write token.


MikeMcQuaid (approved):
Good idea, thanks again @Bo98!

PR #555, opened 2024-06-17
remove-disabled-packages: add action

bevanjkay:
This PR adds a modified version of the existing `remove-disabled-formulae` action that would work across both `homebrew-core` and `homebrew-cask`, and be useable in a repository that includes both Formula and Casks.

It may be worthwhile adding an additional feature to pass a variable for the amount of months a package should be disabled before being removed, as there may be a difference in process between `homebrew-core` and `homebrew-cask`.

### DIFF
--- a/remove-disabled-packages/action.yml
+++ b/remove-disabled-packages/action.yml
@@ -1,0 +1,17 @@
+name: Remove disabled packages
+description: Delete packages that have a disable date of more than one year ago
+branding:
+  icon: trash-2
+  color: red
+outputs:
+  packages-removed:
+    description: Whether any packages were removed or not
+    value: ${{steps.remove-packages.outputs.packages-removed}}
+runs:
+  using: composite
+  steps:
+    - run: brew ruby "$GITHUB_ACTION_PATH/main.rb"
+      id: remove-packages
+      shell: bash
+      env:
+        HOMEBREW_NO_AUTO_UPDATE: 1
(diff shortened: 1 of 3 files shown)


remove-disabled-packages/action.yml
```diff
@@ -0,0 +1,17 @@
+name: Remove disabled packages
+description: Delete packages that have a disable date of more than one year ago
+branding:
+  icon: trash-2
+  color: red
+outputs:
+  packages-removed:
+    description: Whether any formulae were removed or not
```
MikeMcQuaid, line 8:
```suggestion
    description: Whether any packages were removed or not
```

MikeMcQuaid (approved):
Looks good, thanks @bevanjkay! Could you run `brew style --fix` over the Ruby here, too?

bevanjkay:
This now passes style checks for this directory locally

```
brew style --fix remove-disabled-packages
Inspecting 2 files
..

2 files inspected, no offenses detected
```

and is also still functional, [tested on my own tap](https://github.com/bevanjkay/homebrew-tap/pull/1230).

MikeMcQuaid:
Thanks @bevanjkay!

Bo98:
Can you change homebrew-core to use this and delete `remove-disabled-formulae` entirely? Seems like we're the only user beyond some homebrew-core forks (which are their responsibility to rebase).

PR #557, opened 2024-07-13
setup-homebrew: add brew-gh-api-token setting

woodruffw:
This should default to the current workflow runner's GITHUB_TOKEN, and should not need to be configured by the majority of users.

The primary purpose of this is to pre-configure
`HOMEBREW_GITHUB_API_TOKEN` so that attestations are verified in the CI.

### DIFF
--- a/setup-homebrew/action.yml
+++ b/setup-homebrew/action.yml
@@ -25,6 +25,10 @@ inputs:
     description: Token to be used for GitHub authentication (if using private repositories). This token will persist through other steps for the duration of the job.
     required: false
     default: ""
+  brew-gh-api-token:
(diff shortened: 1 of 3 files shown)


setup-homebrew/action.yml
```diff
@@ -25,6 +25,10 @@ inputs:
     description: Token to be used for GitHub authentication (if using private repositories). This token will persist through other steps for the duration of the job.
     required: false
     default: ""
+  brew-gh-api-token:
```
woodruffw, line 28: NB: The `token` setting above seems to be used exclusively for private repo access, so I added this as a separate setting to avoid widening the former's scope. It also has a default unlike `token`.

MikeMcQuaid (approved):
Looks good, thanks!

PR #558, opened 2024-07-13
bump-packages: better handle action inputs

ZhongRuoyu:
Store the formulae and casks in arrays to eliminate code injection through those inputs.

Also, update the descriptions to note that inputs should be space-separated.

### DIFF
--- a/bump-packages/action.yml
+++ b/bump-packages/action.yml
@@ -9,11 +9,11 @@ inputs:
     description: GitHub token (not the default one)
     required: true
   formulae:
-    description: Formula names, one per line
+    description: Formula names, space-separated
     required: false
     default: ''
   casks:
-    description: Cask names, one per line
+    description: Cask names, space-separated
     required: false
     default: ''
   fork:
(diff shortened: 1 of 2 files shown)


bump-packages/action.yml
```diff
@@ -9,11 +9,11 @@ inputs:
     description: GitHub token (not the default one)
     required: true
   formulae:
-    description: Formula names, one per line
+    description: Formula names, space-separated
     required: false
```
Bo98, line 13: Is this a breaking change or was the description wrong before?

ZhongRuoyu: I believe it was wrong (or rather a misunderstanding of how YAML `>` syntax works).

If it really is one per line then this probably wouldn't work:

https://github.com/Homebrew/actions/blob/250a86a75e69d18d941481f9c02b684363233daf/bump-packages/action.yml#L28

Bo98: Cool, make sense!

woodruffw (approved):

MikeMcQuaid:
Thanks @ZhongRuoyu!

Bo98:
Seeing some bad bumps after this: https://github.com/Homebrew/homebrew-core/pull/177286

ZhongRuoyu:
Whoops, should be fixed by https://github.com/Homebrew/actions/pull/565!

ZhongRuoyu:
Now working as expected: https://github.com/Homebrew/homebrew-core/actions/runs/9927784994/job/27423178212

PR #560, opened 2024-07-13
Revert "Revert "setup-homebrew: add brew-gh-api-token setting""

woodruffw:
Turns out this wasn't the issue with https://github.com/Homebrew/.github/pull/140.

Reverts Homebrew/actions#559

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -123,8 +123,8 @@ if [[ ! "$GITHUB_REPOSITORY" =~ ^.+/(home|linux)brew-core$ ]]; then
     unset HOMEBREW_NO_INSTALL_FROM_API
 fi
 
-# Configure HOMEBREW_GITHUB_API_TOKEN (needed for attestations)
-if [[ -n "${BREW_GH_API_TOKEN}" ]]; then
+# Only do so if the user hasn't already explicitly set HOMEBREW_GITHUB_API_TOKEN.
+if [[ -n "${BREW_GH_API_TOKEN}" && -z "${HOMEBREW_GITHUB_API_TOKEN}" ]]; then
     echo "HOMEBREW_GITHUB_API_TOKEN=${BREW_GH_API_TOKEN}" >> "${GITHUB_ENV}"
 fi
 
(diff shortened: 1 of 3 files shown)


setup-homebrew/main.sh
```diff
@@ -122,6 +123,11 @@ if [[ ! "$GITHUB_REPOSITORY" =~ ^.+/(home|linux)brew-core$ ]]; then
     unset HOMEBREW_NO_INSTALL_FROM_API
 fi
 
+# Configure HOMEBREW_GITHUB_API_TOKEN (needed for attestations)
+if [[ -n "${BREW_GH_API_TOKEN}" ]]; then
+    echo "HOMEBREW_GITHUB_API_TOKEN=${BREW_GH_API_TOKEN}" >> "${GITHUB_ENV}"
```
Bo98, line 128: Will this override workflows that already set `HOMEBREW_GITHUB_API_TOKEN`? Maybe want to check if `HOMEBREW_GITHUB_API_TOKEN` is already set.

woodruffw: Yeah, I'll tweak that, thanks for catching!

woodruffw: PTAL

Bo98: needs to be something like `${HOMEBREW_GITHUB_API_TOKEN-}` to avoid the unbound variable error but LGTM otherwise

setup-homebrew/main.sh
```diff
@@ -122,6 +123,11 @@ if [[ ! "$GITHUB_REPOSITORY" =~ ^.+/(home|linux)brew-core$ ]]; then
     unset HOMEBREW_NO_INSTALL_FROM_API
 fi
 
+# Only do so if the user hasn't already explicitly set HOMEBREW_GITHUB_API_TOKEN.
+if [[ -n "${BREW_GH_API_TOKEN}" && -z "${HOMEBREW_GITHUB_API_TOKEN}" ]]; then
```
woodruffw, line 127:
```suggestion
if [[ -n "${BREW_GH_API_TOKEN}" && -z "${HOMEBREW_GITHUB_API_TOKEN-}" ]]; then
```

MikeMcQuaid (approved):

PR #562, opened 2024-07-13
Handle input strings more robustly

ZhongRuoyu:
Avoid using inputs expanded with the `${{ .. }}` syntax in shell scripts to prevent code injection.

Also:

- `failures-summary-and-bottle-result`: make sure every `printf` gets a format string (`%s`) so that we don't get unexpected format sequences from the input.
- `post-build`: fix use of non-existent input `debug`.

### DIFF
--- a/post-build/action.yml
+++ b/post-build/action.yml
@@ -57,9 +57,12 @@ runs:
     - name: Count bottles
       id: bottles
       if: always()
-      run: '"${GITHUB_ACTION_PATH}/count-bottles.sh" "${{ inputs.debug }}"'
+      run: |
+        "${GITHUB_ACTION_PATH}/count-bottles.sh" "${DEBUG}"
       working-directory: ${{ inputs.bottles-directory }}
       shell: /bin/bash -euo pipefail {0}
+      env:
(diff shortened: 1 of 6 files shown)


post-build/action.yml
```diff
@@ -57,7 +57,8 @@ runs:
     - name: Count bottles
       id: bottles
       if: always()
-      run: '"${GITHUB_ACTION_PATH}/count-bottles.sh" "${{ inputs.debug }}"'
+      run: |
+        "${GITHUB_ACTION_PATH}/count-bottles.sh" ${{ runner.debug == '1' }}
```
woodruffw, line 61: I think there's no injection potential here, but even still it might be nice to put the `${{ .. }}` expansion into an envvar for consistency 🙂

MikeMcQuaid: Also: double check that this fails as expected on a non-zero exit code; somethings moving to using `|` breaks this.

MikeMcQuaid (dismissed):
Agreed with comments, looks good once they are addressed!

woodruffw (approved):

PR #563, opened 2024-07-13
brew-script: remove

ZhongRuoyu:
It's not used in any public repos [^1] and can be easily replaced by a `brew ruby` call.

[^1]: https://github.com/search?q=actions%2Fbrew-script+NOT+is%3Afork&type=code


MikeMcQuaid (approved):
Makes sense! That said: we should reduce our uses of `brew ruby` where possible or e.g. figure out how to warn loudly on non-public API usage 😁

PR #566, opened 2024-07-14
workflows/vendor-node-modules: eliminate use of `pull_request_target`

ZhongRuoyu:
This makes it no longer work for PRs from forks, but maintainers can still trigger it via `workflow_dispatch` after verifying the PR contents.


MikeMcQuaid (approved):

woodruffw (approved):

MikeMcQuaid:
Thanks @ZhongRuoyu!

PR #573, opened 2024-08-03
pre-build: set HOMEBREW_NO_UPDATE_REPORT_NEW

Bo98:
The new formula list is very noisy when the images haven't been updated in a while and is not useful information at all for Homebrew/core CI purposes.


p-linnane (approved):

ZhongRuoyu (approved):

MikeMcQuaid:
Makes sense, thanks @Bo98!

PR #585, opened 2024-09-11
create-pull-request: add action

Bo98:
A PR creation action that chews at the API limit less than `gh pr create`.


carlocab (dismissed):
We also use the `--reviewer` and `--label` flags of `gh pr create`.

It would be helpful if this action also had the PR# as an `output`.

Bo98:
Added those.

Note that labels + reviewers are an additional API call each. Not sure how valuable `reviewers` is given we auto merge anyway.

carlocab (approved):

PR #587, opened 2024-09-18
create-pull-request: fix empty string handling

ZhongRuoyu:
`core.getInput` returns an empty string if the value is not defined [^1]. Splitting an empty string in JavaScript returns an array with an empty string, which is truthy. That means the checks for labels and reviewers are always true, even if the input is not defined.

```node
> "".split("/")
[ '' ]
> Boolean([""])
true
```

In addition, add log output so that it's easier to figure out what's going on.

[^1]: https://github.com/actions/toolkit/blob/6dd369c0e648ed58d0ead326cf2426906ea86401/packages/core/src/core.ts#L120

### DIFF
--- a/create-pull-request/main.mjs
+++ b/create-pull-request/main.mjs
@@ -29,6 +29,9 @@ async function main() {
     const response = await client.rest.pulls.create(prRequest)
     const prNumber = response.data.number
     const prNodeId = response.data.node_id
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Thanks @ZhongRuoyu!

PR #619, opened 2024-11-30
bump-packages: fix fork input handling

AlternateRT:
This fixes the `fork` input so that the composite actions actually get the `INPUT_FORK` variable.

This was never passed correctly, meaning `brew bump --no-fork ...` is what always ran, even when `fork` was set to `true`.


ZhongRuoyu (approved):
Thanks for the PR @AlternateRT, good catch!

PR #630, opened 2024-12-15
workflows: fix artipacked zizmor findings

p-linnane:
Trying again to fix these. I set `persist-credentials: true` for any jobs that had git operations, and `false` for the others. Let me know if I misunderstood.

### DIFF
--- a/.github/workflows/create-gcloud-instance.yml
+++ b/.github/workflows/create-gcloud-instance.yml
@@ -17,6 +17,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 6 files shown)


Bo98:
> Trying again to fix these. I set `persist-credentials: true` for any jobs that had git operations, and `false` for the others. Let me know if I misunderstood.

Specifically anything that writes to a remote (e.g. git push).

A while back I actually got ahead of the game here and by making most workflows work without any tokens:

* git-user-config just sets a config - no pushing
* setup-commit-signing similar
* git-try-push actually sets up a `file://` remote to push now (it previously did just push to GitHub)

So if I did it correctly, you shouldn't need credentials for anything here anymore.

Bo98 (approved):

PR #645, opened 2025-02-03
setup-commit-signing: move to SSH signing

woodruffw:
This adds an SSH signing mode to the `setup-commit-signing` action.

This mode has a few advantages:

1. It doesn't require us to use GPG/assume a functional GPG client and agent on the runner. We also get ed25519 instead of a very long RSA key.
2. Provisioning an SSH signing key into the runner is much easier: configuring the key for use with `git` is ~20 lines of straightforward shell.

For now, the mode is still GPG by default, and SSH signing can be enabled by passing `ssh: true` to the action instead. This default will flip in the near future, followed shortly by removing the GPG pathways entirely.

Some things that need to be done before this gets rolled out:

- [ ] We need to configure the new Homebrew org-level SSH signing key.
- [ ] Begin updating uses to do SSH signing instead of GPG signing.


.github/workflows/setup-commit-signing.yml
```diff
@@ -32,43 +32,22 @@ jobs:
...
-            Name-Email: ${GIT_EMAIL}
-            Passphrase: ${PASSPHRASE}
-            Expire-Date: seconds=900
-          EOS
-          KEY=$(echo "${PASSPHRASE}" | gpg --batch --pinentry-mode=loopback --passphrase-fd=0 --armor --export-secret-subkeys "<${GIT_EMAIL}>" | base64 --wrap=0)
-          echo "::add-mask::${KEY}"
-          echo "::add-mask::${PASSPHRASE}"
+          ssh-keygen -t ed25519 -C "Test CI key" -f /tmp/test-ssh-key -N ""
+          KEY="$(cat /tmp/test-ssh-key)"
+          echo "${KEY}" | sed 's/^ */::add-mask::/' # shellcheck disable=SC2001
```
woodruffw, line 41: NB: This `sed` can't be rewritten as a `${foo//find/replace}` expr, so I've disabled the rule for the offending line here.

woodruffw:
I realized it'll be pretty disruptive to make this change in-place, so I'm going to make a new `ssh:` input and then flip the default after we've had the chance to migrate all uses.

setup-commit-signing/action.yml
```diff
@@ -1,19 +1,46 @@
...
   using: composite
   steps:
-    - run: |
+    - name: Enable SSH agent
+      if: inputs.ssh == 'true'
+      run: |
+        eval $(ssh-agent)
+
+        echo "SSH_AUTH_SOCK=${SSH_AUTH_SOCK}" >> ${GITHUB_ENV} # zizmor: ignore[github-env]
+        echo "SSH_AGENT_PID=${SSH_AGENT_PID}" >> ${GITHUB_ENV} # zizmor: ignore[github-env]
```
woodruffw, line 24: NB: This `GITHUB_ENV` writes are safe always safe (single-line), assuming `ssh-agent` itself produces the expected output.

p-linnane (approved):

Bo98 (approved):

MikeMcQuaid (commented):
Lovely stuff!

PR #654, opened 2025-02-19
setup-commit-signing: remove GPG signing support

woodruffw:
This tears out the GPG flows, leaving only the SSH signing flow. This should be safe to do now that we've fully switched to our SSH signing key in all Homebrew org repos.

CC @p-linnane: I think with this we can delete (and optionally revoke) the old GPG secrets as well 🙂


p-linnane (approved):

p-linnane:
Thanks for this! I'd like to give this a week or so post-merge before revoking and deleting.

PR #717, opened 2025-07-08
Enable GHES support in `setup-homebrew`

### LINKED ISSUE #716: Adjust the actions so they could be runned on GitHub Enterprise Servers
Replace `github.com` references in actions which reference to the repository by the actual server url.

e.g.:
https://github.com/Homebrew/actions/blob/94f27a5b7e8d55b5ad5e341889f84ad25bd49727/git-user-config/main.js#L16

https://github.com/Homebrew/actions/blob/94f27a5b7e8d55b5ad5e341889f84ad25bd49727/git-try-push/main.js#L44

Paulchen5:
As first step to resolve #716 I replaced the `github.com` references which are probably pointing to the own tap repositories by this `$GITHUB_SERVER_URL`.

### DIFF
--- a/setup-homebrew/main.sh
+++ b/setup-homebrew/main.sh
@@ -209,12 +209,12 @@ else
 
         if [[ -d "$HOMEBREW_TAP_REPOSITORY" ]]; then
             cd "$HOMEBREW_TAP_REPOSITORY"
-            git remote set-url origin "https://github.com/$GITHUB_REPOSITORY"
+            git remote set-url origin "${GITHUB_SERVER_URL}/$GITHUB_REPOSITORY"
         else
             mkdir -vp "$HOMEBREW_TAP_REPOSITORY"
             cd "$HOMEBREW_TAP_REPOSITORY"
             git init
-            git remote add origin "https://github.com/$GITHUB_REPOSITORY"
+            git remote add origin "${GITHUB_SERVER_URL}/$GITHUB_REPOSITORY"
         fi
 
         # Make repo available under `GITHUB_WORKSPACE` (default working directory), which some third-party taps may need.


Paulchen5:
Can you give a more detailed example how I can test the changes local / by myself?

Do you want me to add the other enablements also in this pr or should I open another pr?

Thanks in advance 👍

setup-homebrew/main.sh
```diff
@@ -136,7 +136,7 @@ fi
 ohai "Fetching Homebrew/brew..."
 if [[ "$GITHUB_REPOSITORY" =~ ^.+/brew$ ]]; then
     cd "$HOMEBREW_REPOSITORY"
-    git remote set-url origin "https://github.com/$GITHUB_REPOSITORY"
+    git remote set-url origin "https://${GITHUB_SERVER_URL}/$GITHUB_REPOSITORY"
```
Bo98, line 139: The documentation states `GITHUB_SERVER_URL` includes `https://`: https://docs.github.com/en/actions/reference/variables-reference

```suggestion
    git remote set-url origin "${GITHUB_SERVER_URL}/$GITHUB_REPOSITORY"
```

Bo98 (approved):

Bo98:
> Can you give a more detailed example how I can test the changes local / by myself?

For Actions involving JavaScript, `npm test` usually covers it. `setup-homebrew` is different though. The best way to test that is to create a tap on your GHES instance and use your fork to test it, e.g.

```yml
- uses: Paulchen5/homebrew-actions@enable-ghes-support
```

> Do you want me to add the other enablements also in this pr or should I open another pr?

Doesn't really matter but let's just do `setup-homebrew` first and then up to you whether you do one PR for the rest or multiple.

Paulchen5:
> The best way to setup-homebrew is to create a tap on your GHES instance and use your fork to test it, e.g.

This wont be possible to some restrictions as long as this is not updated here. Afterwards there will be the approval to use / test the GH action on the Enterprise Server 😅

Paulchen5:
> Worth noting my suggestion above only covered one line - you'll need to change the others too.

oh alr 👍

node_modules/.package-lock.json
```diff
@@ -21,14 +21,18 @@
       }
     },
     "node_modules/@actions/github": {
-      "version": "6.0.0",
-      "resolved": "https://registry.npmjs.org/@actions/github/-/github-6.0.0.tgz",
-      "integrity": "sha512-alScpSVnYmjNEXboZjarjukQEzgCRmjMv6Xj47fsdnqGS73bjJNDpiiXmp8jr0UZLdUB6d9jW63IcmddUP+l0g==",
+      "version": "6.0.1",
```
SMillerDev, line 24: Why change this?

node_modules/@actions/github/lib/context.d.ts
```diff
@@ -11,6 +11,7 @@ export declare class Context {
     action: string;
     actor: string;
     job: string;
+    runAttempt: number;
```
SMillerDev, line 14: Why change this?

node_modules/.package-lock.json
SMillerDev: Please remove it, it makes the actual changes harder to review.

Paulchen5:
Is brew preinstalled on the GitHub runners?


The following check fails at a self-hosted runner:
```bash
# Check brew's existence
if ! which brew &>/dev/null; then
    PATH="/home/linuxbrew/.linuxbrew/bin:/opt/homebrew/bin:/usr/local/bin:$PATH"

    if ! which brew &>/dev/null; then
        echo "Could not find 'brew' command in PATH or standard locations."
        exit 1
    fi
fi
```

SMillerDev:
Yes it is

setup-homebrew/main.sh
```diff
@@ -136,7 +138,7 @@ fi
 ohai "Fetching Homebrew/brew..."
 if [[ "$GITHUB_REPOSITORY" =~ ^.+/brew$ ]]; then
     cd "$HOMEBREW_REPOSITORY"
-    git remote set-url origin "https://github.com/$GITHUB_REPOSITORY"
+    git remote set-url origin "${GITHUB_SERVER_URL}/$GITHUB_REPOSITORY"
```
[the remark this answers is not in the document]
Paulchen5: > The default environment variables that GitHub sets are available to every step in a workflow.
from https://docs.github.com/en/actions/reference/variables-reference#default-environment-variables

So it should be the case 🤷‍♂️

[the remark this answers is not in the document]
MikeMcQuaid: Great, thanks for reference.

setup-homebrew/main.sh
```diff
@@ -209,12 +211,12 @@ else
...
         if [[ -d "$HOMEBREW_TAP_REPOSITORY" ]]; then
             cd "$HOMEBREW_TAP_REPOSITORY"
-            git remote set-url origin "https://github.com/$GITHUB_REPOSITORY"
+            git remote set-url origin "${GITHUB_SERVER_URL}/$GITHUB_REPOSITORY"
         else
             mkdir -vp "$HOMEBREW_TAP_REPOSITORY"
             cd "$HOMEBREW_TAP_REPOSITORY"
             git init
-            git remote add origin "https://github.com/$GITHUB_REPOSITORY"
+            git remote add origin "${GITHUB_SERVER_URL}/$GITHUB_REPOSITORY"
```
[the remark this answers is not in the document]
Paulchen5: I guess I adjusted it according to your requests.

Paulchen5:
Who is responsible for merging the pr?

PR #762, opened 2025-11-23
label-pull-requests: add patch_content matching for diffs

chenrui333:
add patch_content matching for diffs

- https://github.com/Homebrew/homebrew-core/pull/255641

### DIFF
--- a/label-pull-requests/main.mjs
+++ b/label-pull-requests/main.mjs
@@ -46,6 +46,12 @@ async function main() {
 
         // Extend every file object with its content
         for (const file of files.data) {
+            // Only match added/removed lines
+            file.patch = (file.patch || "")
+                .split("\n")
+                .filter(line => /^[-+]/.test(line))
+                .join("\n")
+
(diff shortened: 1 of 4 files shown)


label-pull-requests/main.mjs
```diff
@@ -46,6 +46,8 @@ async function main() {
 
         // Extend every file object with its content
         for (const file of files.data) {
+            file.patch ||= ""
+
```
branchvincent: Would be nice if callers didn't need to include `^[+-]` since they always want to look at changed lines (otherwise they would just use `contents`) and instead handle that here, what do you think? Maybe something like

```suggestion
            file.patch ||= ""
            file.patch = file.patch
                .split("\n")
                .filter(line => /^[-+]/.test(line))
                .join("\n");
```

MikeMcQuaid (approved):
Looks good when comments addressed!

branchvincent:
Verified in https://github.com/Homebrew/homebrew-core/pull/255964

PR #765, opened 2025-11-27
node_modules: remove dependency CI workflow file

cho-m:
`esm-reload` package's CI workflow file is causing some alerts so maybe we can just remove it.

---

In future, should replace `node_modules` with an alternative as recommended by current documentation (single file via rollup.js or ncc):
- https://docs.github.com/en/actions/tutorials/create-actions/create-a-javascript-action#commit-tag-and-push-your-action


woodruffw (approved):

MikeMcQuaid (approved):
Thanks!

PR #770, opened 2025-12-09
Add setup-homebrew support for ubuntu-slim.

MikeMcQuaid:
This will allow it to be installed and used without a pre-existing Homebrew installation.


p-linnane (approved):
Let's give it a shot!

PR #811, opened 2026-03-12
Update action runtimes to node24

Goooler:
Two actions were still on `node20` while the rest of the repo had already migrated to `node24`.

- **`setup-homebrew/action.yml`**: `using: node20` → `using: node24`
- **`wait-for-idle-runner/action.yml`**: `using: node20` → `using: node24`
- **`.github/workflows/vendor-node-modules.yml`**: `node-version: '20'` → `node-version: '24'`

https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/setup-homebrew/action.yml
+++ b/setup-homebrew/action.yml
@@ -37,6 +37,7 @@ outputs:
   gems-hash:
     description: Homebrew's Ruby Gemfile.lock sha256sum
 runs:
-  using: node24
+  # TODO: ask @Bo98 whether migration is complete before updating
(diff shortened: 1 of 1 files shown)


Bo98:
`setup-homebrew` is intentional right now as we need to update one of our CI images which will happen in the next few days (was delayed from Monday).
`wait-for-idle-runner` was indeed missed by accident however, though as a part of the Node 24 updates we have been updating our actions to use TypeScript at the same time (`.mts`).

PR #838, opened 2026-04-06
Warn `@master` users to migrate to `@main`

MikeMcQuaid:
- External workflows still reference `@master`; GitHub Actions does not auto-redirect after a branch rename
- The `master` branch sync will stop and this warning will become an error in Homebrew 5.2.0 (no earlier than 2026-06-10)
- Shared `deprecate-master.{mts,sh}` keep the message in one place across all 23 actions

See also https://github.com/Homebrew/brew/pull/21937

Code written with Claude Claude, several local review and edit passes.

### DIFF
--- a/deprecate-master.sh
+++ b/deprecate-master.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+# Warn users referencing Homebrew/actions via @master to migrate to @main.
+# Called from composite actions with the action name as the sole argument.
+
+set -euo pipefail
+
+if [ -z "${GITHUB_ACTION_REF+x}" ]; then
+  echo "::error::GITHUB_ACTION_REF is not set. This script must be run inside a GitHub Actions workflow."
+  exit 1
+fi
+
+if [ "$GITHUB_ACTION_REF" = "master" ]; then
+  echo "::warning::Homebrew/actions/${1}@master is deprecated. Please update your workflow references to use Homebrew/actions/${1}@main. The 'master' branch sync will stop and this warning will become an error when Homebrew 5.2.0 is released (no earlier than 2026-06-10)."
+fi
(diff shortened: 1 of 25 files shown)


deprecate-master.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# Warn users referencing Homebrew/actions via @master to migrate to @main.
+# Called from composite actions with the action name as the sole argument.
+
+set -euo pipefail
+
+if [ "${GITHUB_ACTION_REF:-}" = "master" ]; then
```
Bo98: can we make this required? I'm worried we might hit https://github.com/actions/runner/issues/2473 and it would be good for CI to confirm whether this is set or not.

Bo98 (approved):

deprecate-master.sh
MikeMcQuaid: @Bo98 I don't understand what "make this required" means. Can you elaborate and/or suggest some code?

Bo98: I mean it would be good to fail if `GITHUB_ACTION_REF` is not set

```suggestion
if [ "${GITHUB_ACTION_REF}" = "master" ]; then
```

Bo98: If CI fails with the above then we are hitting the GitHub Actions bug and we might need our own env and to do something like:

```yaml
env:
  ACTION_REF: ${{ github.action_ref }}
```

Though be aware of the [other bug](https://github.com/actions/runner/issues/2473#issuecomment-1776029708) so may have to use inputs instead.

MikeMcQuaid: @Bo98 Done, not exactly how you suggested but more or less.

Bo98 (approved):

PR #850, opened 2026-05-12
Tweak setup-ruby bundler cache

MikeMcQuaid:
- Keep `bundle install` available without `bundler-cache` so Gemfile projects still verify dependencies.
- Key cached gems by Ruby version and resolved portable Ruby mode to avoid reusing incompatible installs.
- Preserve `bundle check` so warm dependency sets skip needless installs.


SMillerDev (approved):
Looks okay

PR #861, opened 2026-06-01
Trust taps on stable brew

MikeMcQuaid:
- Extends tap trust setup once stable `brew` exposes `brew trust`.
- Lets stable tap CI use the same trust path as non-stable `brew`.


p-linnane (approved):
🎉

PR #873, opened 2026-06-19
bump-packages: support tap-qualified package names

chenrui333:
Summary:
- Auto-enable `--full-name` for formula and cask names containing `/`.
- Remove the redundant `full-name` input.
- Update docs and tests for both command paths.

### DIFF
--- a/bump-packages/action.yml
+++ b/bump-packages/action.yml
@@ -43,7 +43,7 @@ runs:
         fi
         args+=(--open-pr)
         IFS=" " read -r -a formulae <<<"${INPUT_FORMULAE}"
-        brew "${args[@]}" --formulae "${formulae[@]}"
+        brew bump "${args[@]}" --formulae "${formulae[@]}"
       shell: bash
       if: inputs.formulae != ''
       env:
--- a/bump-packages/action.yml
+++ b/bump-packages/action.yml
@@ -20,10 +20,6 @@ inputs:
     description: Use a fork when opening a PR
     required: false
     default: ${{ github.repository_owner != 'Homebrew' }}
-  full-name:
-    description: Force fully-qualified package names to brew bump
-    required: false
-    default: 'false'
 runs:
   using: composite
   steps:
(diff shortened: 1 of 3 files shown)


bump-packages/action.yml
```diff
@@ -30,33 +34,41 @@ runs:
...
+        if [[ "$INPUT_FORK" != "true" ]]; then
+          args+=(--no-fork)
+        fi
+        if [[ "$INPUT_FULL_NAME" == "true" ]]; then
+          args+=(--full-name)
         fi
+        args+=(--open-pr)
         IFS=" " read -r -a formulae <<<"${INPUT_FORMULAE}"
-        brew bump $NO_FORK --open-pr --formulae "${formulae[@]}"
+        brew "${args[@]}" --formulae "${formulae[@]}"
```
carlocab:
```suggestion
        brew bump "${args[@]}" --formulae "${formulae[@]}"
```
I'd keep `bump` here instead of sticking it in `args`. Optional.

carlocab (approved):
Looks fine; optional nit.

MikeMcQuaid (commented):
Looks good! Main question: what would be downsides of making full-name the default behaviour or at least automatic if e.g. a name contains `/`?

chenrui333:
Yeah, I think we turn on the auto detection behavior, just pushed a change to auto-enable `--full-name` when package names contain `/`, which should save action config while preserving existing short-name behavior.

bump-packages/action.yml
```diff
@@ -20,6 +20,10 @@ inputs:
     description: Use a fork when opening a PR
     required: false
     default: ${{ github.repository_owner != 'Homebrew' }}
+  full-name:
+    description: Force fully-qualified package names to brew bump
+    required: false
+    default: 'false'
```
MikeMcQuaid: If this is now auto-detected: do we need this option/input at all?

MikeMcQuaid (commented):
Thanks! One more tweak

MikeMcQuaid (approved):
Thanks!

PR #883, opened 2026-07-10
release: add repository-wide CalVer versioning

p-linnane:
Release all the Actions in this repo together under bare CalVer tags (`YYYY.MM.DD.N`, where `N` restarts at 1 each UTC day), the same lightweight-tag + `gh release create` model `Homebrew/brew` uses. The one difference is that we pin `--target` to the exact `main` SHA captured at checkout instead of trusting whatever the default branch points at when the release publishes.

A scheduled workflow runs every Monday at 12:00 UTC and can also be dispatched manually for an urgent fix. It skips when nothing has changed since the latest CalVer tag; scheduled runs additionally wait for the candidate commit to bake for 24h (manual dispatch skips that wait but still won't cut a no-op release). GitHub creates the tag and generated notes; existing tags are never moved or reused.

The six internal `Homebrew/actions/…@main` references in `pre-build`, `setup-ruby`, `bump-formulae`, and `post-build` are now pinned to full SHAs so a released snapshot never resolves `@main` at runtime. Two tests keep those pins honest: one rejects any internal dependency that isn't a 40-char SHA, the other fails if a pin falls behind its sub-action's current tree. `test.yml` fetches full history so that freshness check can resolve the pinned SHAs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,59 @@
+name: Release
+
+on:
+  schedule:
+    # Every Monday at 12:00 UTC.
+    - cron: "0 12 * * 1"
+  workflow_dispatch:
+
+permissions: {}
+
+defaults:
+  run:
+    shell: bash -euo pipefail {0}
+
+concurrency:
+  group: release
+  cancel-in-progress: false
+
+jobs:
+  release:
+    name: Release
+    if: github.repository_owner == 'Homebrew'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # for creating the release and tag
+    steps:
+      - name: Check out main
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+          persist-credentials: false
+          ref: main
+
+      - name: Prepare release
+        id: release
+        env:
+          RELEASE_MINIMUM_AGE_SECONDS: 86400
+        run: .github/scripts/prepare-release.sh
+
+      - name: Create release
+        if: steps.release.outputs.release == 'true'
+        env:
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -4,6 +4,26 @@ Navigate to subdirectories for more information about particular Actions.
 
 Every directory (except `node_modules`) is a separate Action.
 
+## Releases
+
+All Actions in this repository are released together using bare CalVer tags in
+the form `YYYY.MM.DD.N`, where `N` starts at 1 each UTC day. Release tags point
```
MikeMcQuaid, line 10: I think nicer would be:
```suggestion
the form `YYYY.MM.DD` or `YYYY.MM.DD.N` where N starts at 1 each UTC day if there are multiple releases in a day. Release tags point
```
or

```suggestion
the form `YYYY.[M]MDD.N`, where `N` starts at 0 each UTC day. Release tags point
```
so it's more semver like a.b.c.

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+          persist-credentials: false
+          ref: main
+
+      - name: Prepare release
+        id: release
+        env:
+          RELEASE_MINIMUM_AGE_SECONDS: 86400
```
MikeMcQuaid, line 40: Might want to explain a bit more what this does in docs

MikeMcQuaid (approved):
Great work! Few small nits, feel free to address and self-merge.

README.md
p-linnane: I’d prefer that cleaner format as well, but Dependabot struggles with mixed-arity versions, which can leave existing pins stranded. Keeping all tags four-part avoids that.

p-linnane:
zizmor findings are because `main` has moved. They need to be updated once I cut the first tag anyway, so resolving and merging.

PR #902, opened 2026-07-31
setup-homebrew: output tap name

MikeMcQuaid:
Expose the already-computed `HOMEBREW_TAP_NAME` as a `tap-name` output so tap workflows, e.g. the `brew tap-new` autobump workflow from Homebrew/brew#23349, can use it instead of hardcoding it.

Closes https://github.com/Homebrew/actions/pull/904

### DIFF
--- a/release.test.mts
+++ b/release.test.mts
@@ -248,11 +248,13 @@ test("internal Homebrew/actions dependencies use full commit SHAs", () => {
   assert.deepEqual(invalidReferences, []);
 });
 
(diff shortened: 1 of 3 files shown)


ZhongRuoyu (approved):
Makes sense!

SMillerDev (approved):

SMillerDev:
@MikeMcQuaid I think this needs a rebase after the dependabot merge?

PR #903, opened 2026-07-31
Use self-repository syntax for internal actions

MikeMcQuaid:
Blocked on https://github.com/zizmorcore/zizmor/issues/2248
Blocked on all Homebrew self-hosted runners being on runner version ≥ 2.336.0

- GitHub now resolves `$/name` references to this repository at the commit that is running, so internal references can no longer lag behind the latest release or require a checkout: https://github.blog/changelog/2026-07-30-reference-same-repository-actions-with-self-repository-syntax/
- Replace pinned `Homebrew/actions/*` and checked-out `./*` internal references in composite actions and workflows with `$/*`.
- Leave workflows synced from the `.github` repository unchanged as they run in other repositories where `$/` resolves elsewhere.
- Drop the checkout that only served `./git-user-config/` in its test.
- Replace the release test SHA pin checks, which `$/` obsoletes, with checks that internal references use `$/` and target real actions.
- Ignore actionlint's invalid-format error for `$/` references until actionlint understands the new syntax.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -5,3 +5,9 @@ self-hosted-runner:
 # organization. `null` means disabling configuration variables check.
 # Empty array means no configuration variable is allowed.
 config-variables: []
+paths:
+  .github/workflows/**/*.yml:
+    ignore:
+      # actionlint does not yet understand the self-repository (`$/`) syntax:
+      # https://github.blog/changelog/2026-07-30-reference-same-repository-actions-with-self-repository-syntax/
+      - 'specifying action "\$/[^"]+" in invalid format because ref is missing'
--- a/.github/workflows/cache-homebrew-prefix.yml
+++ b/.github/workflows/cache-homebrew-prefix.yml
@@ -20,7 +20,7 @@ jobs:
           persist-credentials: false
 
       - name: Set up Homebrew
-        uses: ./setup-homebrew/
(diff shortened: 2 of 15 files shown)


ZhongRuoyu (approved):

p-linnane (approved):
Our self-hosted macOS runners are on 2.336.0.

Bo98:
GitHub UI isn't showing this very well but:

> The action $/setup-homebrew is not allowed in Homebrew/actions because all actions must be from a repository owned by Homebrew, created by GitHub, or match one of the patterns: 1Password/load-secrets-action@\*, aws-actions/amazon-ecr-login@\*, aws-actions/amazon-ecs-deploy-task-definition@\*, aws-actions/amazon-ecs-render-task-definition@\*, aws-actions/configure-aws-credentials@\*, bridgecrewio/checkov-action@\*, codecov/codecov-action@\*, docker/build-push-action@\*, docker/login-action@\*, docker/setup-buildx-action@\*, google-github-actions/setup-gcloud@\*, opentofu/setup-opentofu@\*, ruby/setup-ruby@\*, rubygems/configure-rubygems-credentials@\*, rubygems/release-gem@\*. All actions must also be pinned to a full-length commit SHA.

Unless they fixed it in the last few days since the last run, it seems to be buggy in that it thinks `$` isn't `Homebrew`.

PR #905, opened 2026-07-31
Replace commit status with workflow result

MikeMcQuaid:
- Fail validation through the workflow for read-only fork tokens.
- Preserve existing package-repository label management.
- Document Issues permissions and branch protection migration.

### DIFF
--- a/check-commit-format/README.md
+++ b/check-commit-format/README.md
@@ -7,11 +7,33 @@ commits follow [Homebrew's commit style guidelines](https://docs.brew.sh/Formula
 ## Usage
 
 ```yaml
-- name: Check commit style
-  uses: Homebrew/actions/check-commit-format@1f8e202ffddf94def7f42f6fa3a482e821489f9c # 2026.07.10.1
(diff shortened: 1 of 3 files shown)


andrew (approved):
Looks good, this unblocks fork PRs on Homebrew/brew (and is the missing half of Homebrew/brew#23402, which currently pins a SHA that no longer exists here).

One non-blocking note for the package-repo path: at `main.mts:207-215`, `issues.update` runs before `core.setFailed(failure_message)`. Fork `pull_request` tokens get `issues: write` downgraded to read (same as statuses), so if the label update throws, the outer `catch` reports the HTTP error and the validation message is lost. Previously the status was posted before this call so the message survived. Calling `core.setFailed(failure_message)` before the label block, or wrapping the update in `try/catch` with `core.warning`, would keep the useful message. Doesn't affect the `check_package_commit_format: false` case brew uses, so not blocking.

p-linnane (approved):

PR #906, opened 2026-08-04
setup-homebrew: remove Bubblewrap logic

MikeMcQuaid:
- Homebrew's Linux sandbox has moved to Landlock, which needs no helper binary, kernel sysctls or `--privileged` containers.
- Remove the `setup-sandbox` input whose only job was installing and configuring Bubblewrap via `brew setup-sandbox`.
- Rely on Landlock as the default implementation instead of setting `HOMEBREW_SANDBOX_LINUX_LANDLOCK` in CI.

Needs https://github.com/Homebrew/brew/pull/23425

### DIFF
--- a/.github/workflows/setup-homebrew.yml
+++ b/.github/workflows/setup-homebrew.yml
@@ -31,16 +31,18 @@ jobs:
           - os: ubuntu-latest
             key: docker-root
             stable: false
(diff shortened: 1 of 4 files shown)


p-linnane (approved):
🚀

PR #926, opened 2026-09-10
remove-disabled-packages: add tests

p-linnane:
`remove-disabled-packages` had no unit tests. This runs the action's Ruby step against a stubbed `brew` and a temporary Git tap, covering the twelve month removal cutoff on both sides, cask removal and the `Casks` directory gate, filtering packages from other taps, `ruby_source_path` validation, and UTF-8 `brew info` output under `LC_ALL=C`.

Testing this surfaced a bug: `main.rb` called `Pathname()` without requiring `pathname`, which only works because Ruby 4.0 loads it transitively through `fileutils` and `open3`. It raises `NoMethodError` on 3.2 through 3.4.

The Node test job moves to `ubuntu-latest` because `ubuntu-slim` does not ship Ruby. Its Ruby is older than the portable Ruby the action uses in production, which is what caught the missing require; `.github/workflows/remove-disabled-packages.yml` continues to exercise the action end to end on portable Ruby.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ permissions:
 
 jobs:
   test:
-    runs-on: ubuntu-slim
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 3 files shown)


MikeMcQuaid (approved):
Thanks!

PR #929, opened 2026-09-12
create-or-update-issue: add type

cho-m:
Trying to replace `bug` labels with type.

Not sure if this actually works as untested.


p-linnane (approved):

MikeMcQuaid (approved):
Happy to give it a try!

